PR #3, opened 2017-05-27
Create README.md

antoniovj1:
I think it is useful to have a readme for easy access to the development information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,42 @@
+# MyTourBook
+
+## [Official Website](http://mytourbook.sourceforge.net/mytourbook/)
+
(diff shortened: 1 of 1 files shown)


wolfgang-ch:
Thanks for your contribution, there are few things which could be improved:
- What I'm missing is that the data directory is different for [Linux](http://mytourbook.sourceforge.net/mytourbook/index.php/download-install/data-directory)
- The list _Others things..._ is just a few of many links which are important for developer

wolfgang-ch:
I created a readme.md file but without repeating text from the project website

PR #28, opened 2019-01-07
When selecting a tour with no GPS data, the previous selected activity

FJBDev:
should not appear. In this case, we display the default position


wolfgang-ch:
When your code would be merged and I have **not** selected a sync tour or sync slider action, then the map is still moved to the default location. 

Sometimes I disable these sync actions that I can see the selected tours but don't want that the map is moved. 

The method showDefaultMap() should not move the map because it is called in different situations.

A solution for your situation could be that a new map option/action can decide what to do.

FJBDev:
Ok, I will investigate further to find a better solution

FJBDev:
@wolfgang-ch : Here is another solution where we move to the default location ONLY when the **current selected tour has no GPS data.**

wolfgang-ch:
The map must have the ability, whenever a tour (of all kinds) is selected, the map is not relocated.

As in my previous post pointed out, there could be an option in the map options slideout to enable/disable the feature which you have implemented

FJBDev:
Sure, my initial goal was not to move the map but so far I can't find a solution to hide the previous track and stay at the same position. The showDefaultMap() doesn't seem to clear the previous displayed track.

I will continue to investigate more.

FJBDev:
Here is another way of handling the original issue :

FJBDev:
I have just tested and it works well, that solution is great!

PR #39, opened 2019-03-31
Map25: switching mapbookmark clustering on/off via gui

telemaxx:


wolfgang-ch:
Can you convert

public int MapBookmarkEventType_moveto = 0;
public int MapBookmarkEventType_modified = 1;

in 

https://github.com/wolfgang-ch/mytourbook/blob/map25/bundles/net.tourbook/src/net/tourbook/map/bookmark/IMapBookmarks.java#L22-L23

into an Enum.

In my early Java days, I used also this way to define types but an Enum is more type save

wolfgang-ch:
This is an example how I use an enum in a similar case

https://github.com/wolfgang-ch/mytourbook/blob/map25/bundles/net.tourbook/src/net/tourbook/tour/TourEventId.java

telemaxx:
some enums in the next PR...

here you are also using enums:
net.tourbook.map25.TileEncoding

PR #43, opened 2019-04-12
Fix NPE when time is adjusted in the tour filter

Mkhozi:


wolfgang-ch:
Which code-page are you using in the editor?

I'm using UTF-8

mt-review-code-page

wolfgang-ch:
I think it would look better when all controls of the 1st line for the between date/times are aligned vertically to the top as the word **and** is already aligned

mt-review-tour-filter

PR #44, opened 2019-04-13
Map25: removed debug code and new MapBookmarkCluster Icon

telemaxx:


telemaxx:
that was easy to add an additional change to my existing PR. Cool

PR #48, opened 2019-04-18
 Importing the wind speed from ST 

FJBDev:


FJBDev:
I see you later did a commit "reformatted code to the current layout settings "
I'm not sure why you had to do this as I use your code formatter profile. Any ideas ?

image

wolfgang-ch:
I don't know which version you are using but I discovered that you are still using tabs instead of spaces.

mt-spacing

This is the history of the formatter
mt-formatting



I also adjusted the imports sequence to be sorted alphabetically, all formatting setting are described here
http://mytourbook.sourceforge.net/mytourbook/index.php/development/dev-ide/code-style

For me, the formatting is a continuous adjustment when new features are available or when I don't like something in the old format, e.g. when code lines wrap too early. I reformat code only in files which I'm changing.

FJBDev:
I have the same history
image

I will clean up my eclipse and reconfigure it to see if that fixes it.

PR #49, opened 2019-04-23
Suunto9 swimmingdata movesensedata

FJBDev:
- Added support for swimming activities
– Added support for Movesense heart rate belt data
- Fixed a bug where markers/laps would not all be imported
- Removed the import of split files as MTB has fixed the tour concatenation
issue in 19.2.0 (see commit :
https://github.com/wolfgang-ch/mytourbook/commit/b4d5eacfaa4e678273552eb501f39f64ea297956)


FJBDev:
Thanks! Barely made it to 19.4, pfewww ;-)

PR #51, opened 2019-04-26
Fully finished, improved and fixed french translation

FJBDev:


wolfgang-ch:
You should configure the translation tool as described here http://mytourbook.sourceforge.net/mytourbook/index.php/development/translation/install-resource-editor#configure-editor, that the files are correctly formatted, otherwise there are too many differences

mt-translation

FJBDev:
Will do, thanks!

PR #52, opened 2019-04-27
Fixing the case where the reimport filedialog doesn't show

FJBDev:


FJBDev:
On linux platform, this returns null : Display.getDefault().getActiveShell();

Inspired from here : https://stackoverflow.com/questions/22074499/eclipse-rcp-which-shell-to-use-for-popups
this returns a non null value.

wolfgang-ch:
Shell is a beast, I already had to fix some issues when a tooltip/slideout/dialog was not displayed because of the wrong parent shell. 

The most advanced shell in MT is the reparented shell https://github.com/wolfgang-ch/mytourbook/blob/master/bundles/net.tourbook.common/src/net/tourbook/common/tooltip/AbstractRRShell.java

wolfgang-ch:
I've cherry picked these 2 commits from your branch into master but the french translations in your branch are not yet merged because of the formatting

PR #56, opened 2019-05-02
Adding the ability to create a marker from a location on the map

FJBDev:
I would like to add the ability to add a marker from a specific location on the tour from the map view.

To create a marker from a position on the tour. One will have to hover a tour then right click on "Create Marker"

image

The marker is created
image

If the user right clicks on the map without being "on" the tour, the "Create Marker" icon will be disabled as below:
image


telemaxx:
good idea !

wolfgang-ch:
I didn't receive an email notification for this PR, so I saw it just now.

After a tour can now be selected in the 2D map, I also had some ideas for improvements, e.g. create/edit tour marker, but you are faster :-)

1. I tested your PR and saw also in the code, that you use the center of the map to set a tour marker and not the cursor position, I don't think this was intended?

final double clickedTourPointLatitude = this._mapView.getMapLocation().getMapPosition().getLatitude();
final double clickedTourPointLongitude = this._mapView.getMapLocation().getMapPosition().getLongitude();

Your can find the mouse geo position in de.byteholder.geoclipse.map.Map._mouseMove_GeoPosition

2. I would also change the wording **Create Marker** into **Create Tour Marker** as later on there may be other markers

3. Actions are (mostly) enabled/disabled in the method enableActions() all over the app (exceptions do always happen). I would be happy when you move the enablement for the create tour marker action also in this method

FJBDev:
I didn't receive an email notification for this PR, so I saw it just now.
**==>** Weird. I thought it was a one time glitch because I don't know if there is a way to notify you.

After a tour can now be selected in the 2D map, I also had some ideas for improvements, e.g. create/edit tour marker, but you are faster :-)

    I tested your PR and saw also in the code, that you use the center of the map to set a tour marker and not the cursor position, I don't think this was intended?
**==>** Correct, it was not intended. I thought it was giving the location of the mouse click. I am glad you mentioned that because I was finding that sometimes the point found was not accurate. So I will use your recommendation below and probably it will help make the new marker more accurately placed on the map.
I will work on implementing your recommendations, thanks!

final double clickedTourPointLatitude = this._mapView.getMapLocation().getMapPosition().getLatitude();
final double clickedTourPointLongitude = this._mapView.getMapLocation().getMapPosition().getLongitude();

Your can find the mouse geo position in de.byteholder.geoclipse.map.Map._mouseMove_GeoPosition

    I would also change the wording Create Marker into Create Tour Marker as later on there may be other markers

    Actions are (mostly) enabled/disabled in the method enableActions() all over the app (exceptions do always happen). I would be happy when you move the enablement for the create tour marker action also in this method

PR #58, opened 2019-05-07
Modifications per previous PR suggestions

FJBDev:
- Renamed to "Create Tour Marker. "

- Now using the actual mouse click location.

- Improving the user experience by directly showing the new marker location

- Moving the enablement for the create tour marker action in the enableActions() method


wolfgang-ch:
Gmail has put some of the PR notifications into the spam folder which also didn't forward the mails

wolfgang-ch:
Some hints for the wording of the action: **Create tour marker**

* Actions main words start with uppercase as you can see in the other actions within the context menu
* The action has no mnemonic, I would recommend k as this is not yet used

This is my suggestion: **Create Tour Mar&ker**

Can you always merge the last master branch into your branch as it is sometimes difficult (when there are some branches) to find your last anchestor from the master

The mouse cursor position is now very accurate :-)

PR #59, opened 2019-05-09
Modifications per last PR suggestions

FJBDev:
- Renamed Create tour marker to Create Tour Mar&ker
- Merged to the latest master branch

Note : From now on, I will ALWAYS merge the last master branch into my branch


wolfgang-ch:
Now I got an e-mail notification, it seems that the google spam filter has learned

wolfgang-ch:
I just discovered that you save the tour, but what happens when the tour is edited in the tour editor and not yet saved?

For this case there is this method **net.tourbook.tour.TourManager.isTourEditorModified()** which you MUST use, before you create the marker/save tour.

You can see in the other use cases of this method how to use it.

FJBDev:
Thanks, I didn't think about this case.

PR #63, opened 2019-05-13
Small improvements per code review

FJBDev:
- Removing the "jokl" xml element serialized
- Assigning and incrementing the progress bar to a valid value


FJBDev:
@wolfgang-ch Do you want me to add a checkbox in the preferences to ask the user what to do with invalid files.
The checkbox could look like this:
[ ] Save invalid files name & location

wolfgang-ch:
This is a good idea, that this feature could be disabled.

A good location is the root import pref page as this would apply to all importers.

You should also provide a short description below the checkbox (similar to the "Create unique..." checkbox) which should contain the filename with the ignored names.

PR #64, opened 2019-05-16
 adding a checkbox in the preferences to ask the user what to do with  invalid files

FJBDev:
Misc french translation fixes


wolfgang-ch:
You have also modified this german properties file **bundles/net.tourbook/src/net/tourbook/messages_de.properties** and created 2 empty key. Because the keys are created, it cannot be recognized that this needs to be translated.

Before I do a commit, I always check what is modified. Sometimes I see changes which needs to be adjusted.

mt-de-properties

FJBDev:
I think it created the keys automatically. I will remove them

PR #66, opened 2019-05-25
Map25 style names are mostly not localized

telemaxx:


telemaxx:
Problem was that localization string provided by eclipse was like "en_EN", "en" was needed

PR #83, opened 2019-08-03
Adding support for the import of Power data from Garmin RD Pod sensor

FJBDev:


bundles/net.tourbook.device.garmin.fit/src/net/tourbook/device/garmin/fit/listeners/MesgListener_Record.java
```diff
@@ -307,6 +311,11 @@ private void setRecord(final RecordMesg mesg) {
     */
    private void setRecord_DeveloperData(final RecordMesg mesg, final TimeData timeData) {
 
+      final boolean containsMultiplePowerDataSources =
```
wolfgang-ch: `containsMultiplePowerDataSources` should be move to the place where it is used -> below line 358, otherwise it would be computed for any case

PR #88, opened 2019-08-09
Code modifications per Wolfgang's code review

FJBDev:


FJBDev:
> The ration inHG to mb is 1:33, so I think that the air preassure with mb should not be more preciser than with inHG -> 1 digit should be enough, otherwise inHG should have 3 digits

Ok, I will change it to only have 1 decimal when in metric system

FJBDev:
Change performed

bundles/net.tourbook/src/net/tourbook/ui/views/tourDataEditor/TourDataEditorView.java
```diff
@@ -4106,7 +4106,8 @@ private void createUI_Section_147_Weather_Other_Col1(final Composite parent) {
          // spinner: pressure value
          _spinWeather_PressureValue = new Spinner(container, SWT.BORDER);
          _spinWeather_PressureValue.setToolTipText(Messages.Tour_Editor_Label_AirPressure_Tooltip);
-         _spinWeather_PressureValue.setMaximum(400000);
+         //The highest barometric pressure ever recorded on Earth was 32.01 inches (1083.98), measured in Agata, U.S.S.R., on December 31, 1968.
+         _spinWeather_PressureValue.setMaximum(108398);
```
wolfgang-ch: When there would be a new record, it could not be entered into MT

FJBDev: Sorry, I don't understand what you mean. No data cannot be entered ? why ?

wolfgang-ch: You have set the max value for air pressure to 108398 but when there would be a higher value than 108398 (a new record), it could not be entered in this field

FJBDev: Oh, Ok. I will increase the max value.

PR #91, opened 2019-08-15

FJBDev:
Computing the avg and max power values
Setting the isStrideSensorpresent.
Added a new DB field power_DataSource


bundles/net.tourbook.device.garmin.fit/src/net/tourbook/device/garmin/fit/listeners/MesgListener_Session.java
```diff
@@ -105,16 +107,36 @@ public void onMesg(final SessionMesg mesg) {
...
+         final ArrayList<Float> powerDataList = new ArrayList<>();
+         for (final TimeData timeData : fitData.getAllTimeData()) {
+            powerDataList.add(timeData.power);
+
+            if (tourData.getPower_DataSource() == null) {
+               tourData.setPower_DataSource(timeData.powerDataSource);
+            }
+         }
+
+         avgPower = (int) powerDataList.stream().mapToDouble(Float::doubleValue).average().getAsDouble();
```
FJBDev: @wolfgang-ch 
I didn't use what you recommended "tourData.computeAvg_FromValues" as at this stage, the time series not having been calculated, the computeAvg_FromValues would return 0 [here ](https://github.com/wolfgang-ch/mytourbook/blob/5bd78f2e7c88154a49bf6d091b3b6c8ff1cc803f/bundles/net.tourbook/src/net/tourbook/data/TourData.java#L2841)

bundles/net.tourbook/src/net/tourbook/data/TourData.java
```diff
@@ -134,6 +134,8 @@
    public static final int             DB_LENGTH_WEATHER                 = 1000;
    public static final int             DB_LENGTH_WEATHER_CLOUDS          = 255;
 
+   public static final int             DB_LENGTH_POWER_DATA_SOURCE       = 255;
```
FJBDev: @wolfgang-ch 
You said this "I would create the field power_DataSource with variable string length" but I don't know how to specify a variable length in Java. Is 255 ok ?

bundles/net.tourbook/src/net/tourbook/data/TourData.java
```diff
@@ -700,66 +702,57 @@
...
    private int                   rearShiftCount;
 
-   // ############################################# RUNNING DYNAMICS #######################################
+   // ############################################# GEARS #############################################
 
    private short                 runDyn_StanceTime_Min;
    private short                 runDyn_StanceTime_Max;
-   private float                 runDyn_StanceTime_Avg;
 
+   // ############################################# RUNNING DYNAMICS #######################################
```
wolfgang-ch: There are too many formatting modifications

tour-data-reformatted

bundles/net.tourbook/src/net/tourbook/ui/views/tourDataEditor/TourDataEditorView.java
```diff
@@ -4107,7 +4107,7 @@ private void createUI_Section_147_Weather_Other_Col1(final Composite parent) {
          _spinWeather_PressureValue = new Spinner(container, SWT.BORDER);
          _spinWeather_PressureValue.setToolTipText(Messages.Tour_Editor_Label_AirPressure_Tooltip);
          //The highest barometric pressure ever recorded on Earth was 32.01 inches (1083.98), measured in Agata, U.S.S.R., on December 31, 1968.
-         _spinWeather_PressureValue.setMaximum(108398);
+         _spinWeather_PressureValue.setMaximum(110000);
```
wolfgang-ch: This value is OK :-)

bundles/net.tourbook/src/net/tourbook/data/TourData.java
FJBDev: Arg, i dont know why Eclipse does that. I had the same problem with TourDatabase.java. I will fix it

bundles/net.tourbook.device.garmin.fit/src/net/tourbook/device/garmin/fit/listeners/MesgListener_Record.java
```diff
@@ -311,10 +310,14 @@ private void setRecord(final RecordMesg mesg) {
     */
    private void setRecord_DeveloperData(final RecordMesg mesg, final TimeData timeData) {
 
-      final boolean containsMultiplePowerDataSources =
-            StreamSupport.stream(mesg.getDeveloperFields().spliterator(), false)
-                  .anyMatch(developerField -> developerField.getName().equals(DEV_FIELD_NAME__POWER) &&
-                        developerField.getName().equals(DEV_FIELD_NAME__POWER));
+      int powerDataSources = 0;
```
wolfgang-ch: this should be moved below line 361 because it is used only there

bundles/net.tourbook.device.garmin.fit/src/net/tourbook/device/garmin/fit/listeners/MesgListener_Record.java
```diff
@@ -375,6 +378,9 @@ private void setRecord_DeveloperData(final RecordMesg mesg, final TimeData timeD
                }
             }
 
+            timeData.powerDataSource = devField.getName().equals(DEV_FIELD_NAME__POWER) ? Messages.PrefPage_Fit_Combo_Power_Data_Source_Stryd
```
wolfgang-ch: timeData.powerDataSource should contain static text and not text which is translated that later on this field could be used for any kind of reports/statistics/views

bundles/net.tourbook/src/net/tourbook/data/TourData.java
wolfgang-ch: It's always a good idea to review the own changes before committing it

bundles/net.tourbook.device.garmin.fit/src/net/tourbook/device/garmin/fit/listeners/MesgListener_Session.java
wolfgang-ch: this average() method do propably not take into account when you have paused and also the time between 2 time slices, which could be > 10 sec or 1 sec

FJBDev: Correct, but again, I can't use this tourData.computeAvg_FromValues because the time series is null

bundles/net.tourbook.device.garmin.fit/src/net/tourbook/device/garmin/fit/listeners/MesgListener_Record.java
FJBDev: But I thought it would be better to have it out of the for loop so we don't recalculate it every time (more efficient). Thoughts?

bundles/net.tourbook.device.garmin.fit/src/net/tourbook/device/garmin/fit/listeners/MesgListener_Record.java
FJBDev: Resolved in upcoming PR

bundles/net.tourbook.device.garmin.fit/src/net/tourbook/device/garmin/fit/listeners/MesgListener_Session.java
FJBDev: I am looking if I can use the function you advised as you are right, we would want to take into account when the user has paused and also the time between 2 time slices, which could be > 10 sec or 1 sec
If you have a better idea, let me know

FJBDev: Maybe I should set a boolean such as computeAveragePower in the clas FitData.java
and if this boolean is true, I compute the average power [here ](https://github.com/wolfgang-ch/mytourbook/blob/5bd78f2e7c88154a49bf6d091b3b6c8ff1cc803f/bundles/net.tourbook.device.garmin.fit/src/net/tourbook/device/garmin/fit/FitData.java#L176) ??

bundles/net.tourbook.device.garmin.fit/src/net/tourbook/device/garmin/fit/listeners/MesgListener_Record.java
wolfgang-ch: When entering the method, you can check if dev data are available

if (mesg.getDeveloperFields().size() == 0) {
   return
}

bundles/net.tourbook.device.garmin.fit/src/net/tourbook/device/garmin/fit/listeners/MesgListener_Session.java
wolfgang-ch: This is the location where I also have computed the power, after tourdata is filled up

As you can see in the other boolean fields in FitData, they have a prefix _is... which I use to start boolean fields, for me it is easier to read it and I also try to group them together when it makes sense.

FJBDev: Ok, I have renamed it to _isComputeAveragePower

bundles/net.tourbook.device.garmin.fit/src/net/tourbook/device/garmin/fit/listeners/MesgListener_Record.java
FJBDev: Ok, this comment doesn't apply anymore since I have removed this code

wolfgang-ch: The following code is computed for every record in all fit files

fit-record-listener

When putting
```
if (mesg.getDeveloperFields().size() == 0) {
   return;
}
```
at the beginning of the method, any dev field computation would be not necessary

FJBDev: Oh, got it. I misunderstood. I will fix it ASAP.

FJBDev: Done (see upcoming PR)

PR #93, opened 2019-08-19
 Fixes per code review. Adding missing work I did for 19.9 readme

FJBDev:


bundles/net.tourbook.device.garmin.fit/src/net/tourbook/device/garmin/fit/listeners/MesgListener_Record.java
```diff
@@ -310,6 +310,16 @@ private void setRecord(final RecordMesg mesg) {
     */
    private void setRecord_DeveloperData(final RecordMesg mesg, final TimeData timeData) {
 
+      int developerFieldCount = 0;
+      for (@SuppressWarnings("unused")
```
wolfgang-ch: First I though why you do not use mesg.getDeveloperFields().size() because internally an arraylist is used. Now I also discovered, that an interator is returned which has no size() method -> there is not better way to optimize it :-(

The only thing which could be optimized to use a local var for mesg.getDeveloperFields() as it is used 3 times but the compiler may already is doing this.

For me the code in now OK and will merge it into master

FJBDev: I wish I could use .size() but Iterators are more complicated to use than arraylist for example.

PR #95, opened 2019-08-22
 New feature : Possibility to delete marker(s) from the DEL key and the contextual menu

FJBDev:
image


wolfgang-ch:
We should use a common delete icon as it is already done in some locations in MT

I just replaced the delete icon from .gif to .png

delete-icon

The other problem is that when we both use the same .svg file in different branches, this cannot be merged, it could be merged but this is no fun.

FJBDev:
Image fixed (see my PR)
image

wolfgang-ch:
Plural or singular but not two times a **(s)** 

I just found the following rules are very good:

* If there can be maximum 1 entry/item in a given category: Singular

* If there can be more than 1 entry/item in a given category: Plural

found in https://ux.stackexchange.com/questions/110325/plural-or-singular-navigational-menu-options#115236


I know that there can be labels in MT which do not use this rule but now I will try to use them

FJBDev:
> 
> 
> Plural or singular but not two times a **(s)**
> 
> I just found the following rules are very good:
> 
>     * If there can be maximum 1 entry/item in a given category: Singular
> 
>     * If there can be more than 1 entry/item in a given category: Plural
> 
> 
> found in https://ux.stackexchange.com/questions/110325/plural-or-singular-navigational-menu-options#115236
> 
> I know that there can be labels in MT which do not use this rule but now I will try to use them

Ok, I will follow this rule.
In this case, it should be also "Edit markers" with an s right ?
image

wolfgang-ch:
In this case, no. Although you have selected multiple markers but the action can edit only one marker at a time, you need to select another marker in the opened dialog to edit also the other marker.

So the rules needs to be extended that the action has to be done directly to the selected item(s) and no other actions are necessary to do the action in the menu.

bundles/net.tourbook/src/net/tourbook/ui/views/tourMarker/TourMarkerView.java
```diff
@@ -467,6 +472,29 @@ private void createUI_10_TableViewer(final Composite parent) {
...
       table.setLinesVisible(false);
 
+      table.addKeyListener(new KeyListener() {
+
+         @Override
+         public void keyPressed(final KeyEvent e) {
+
+            if (e.keyCode == SWT.DEL) {
+
+               if (_actionDeleteTourMarkers == null) {
```
wolfgang-ch: I think it is not necessary to check agains `null` but it should check if `_actionDeleteTourMarkers` is enabled or not

bundles/net.tourbook/src/net/tourbook/tour/ActionDeleteMarkerDialog.java
```diff
@@ -0,0 +1,132 @@
...
+      String dialogTitle = Messages.Dlg_TourMarker_MsgBox_delete_marker_title;
+      String dialogMessage = NLS.bind(Messages.Dlg_TourMarker_MsgBox_delete_marker_message, (selectedTourMarkers.get(0)).getLabel());
+
+      if (selectedTourMarkers.size() > 1) {
+         dialogTitle = Messages.Dlg_TourMarker_MsgBox_delete_markers_title;
+
+         final StringBuilder markersNames = new StringBuilder();
+         for (final TourMarker tourMarker : selectedTourMarkers) {
+            if (markersNames.toString().isEmpty() == false) {
+               markersNames.append(", "); //$NON-NLS-1$
```
wolfgang-ch: Newlines may be better readable

PR #98, opened 2019-08-27
Adding newlines to be better readable

FJBDev:


bundles/net.tourbook/src/net/tourbook/tour/ActionDeleteMarkerDialog.java
```diff
@@ -76,10 +76,10 @@ private static void doAction(final ITourProvider tourProvider,
       if (selectedTourMarkers.size() > 1) {
          dialogTitle = Messages.Dlg_TourMarker_MsgBox_delete_markers_title;
 
-         final StringBuilder markersNames = new StringBuilder();
+         final StringBuilder markersNames = new StringBuilder(System.getProperty("line.separator")); //$NON-NLS-1$
          for (final TourMarker tourMarker : selectedTourMarkers) {
             if (markersNames.toString().isEmpty() == false) {
-               markersNames.append(", "); //$NON-NLS-1$
+               markersNames.append(System.getProperty("line.separator")); //$NON-NLS-1$
```
wolfgang-ch: When you run MT with a profiler (I got a free lisence for YourKit), then you can see how very often string resources are created for e.g. " ", "\n", ...

Therefore I created constants for these resources, in this case `UI.NEW_LINE`

wolfgang-ch: >Does having a constant make MTB faster ? I've never run a profiler on MTB
I don't know if a constant make it faster but in other cases, e.g. when inlining a method I saw better performance. With Eclipse refactoring you can inline a method, e.g. Rectangle.contains(...)

wolfgang-ch: >Is it normal that NEW_LINE is defined at several places ??

This has historical reasons, in the first few years of MT there was no net.tourbook.common plugin which do not have net.tourbook.* code dependencies and can be used in other plugins

>Also, is it better to use this maybe to have consistent new lines on Windows and Linux ??

In the first place you should use UI from net.tourbook.common, when I see the old code then I replace it when it makes sense (when there is not many other used code from net.tourbook.ui.UI)

bundles/net.tourbook.device/src/net/tourbook/device/sporttracks/FitLogSAXHandler.java
```diff
@@ -337,15 +337,14 @@ private void finalizeTour() {
...
-            tourNotes.append(System.getProperty("line.separator")); //$NON-NLS-1$
+            tourNotes.append(UI.NEW_LINE2);
          }
 
          tourNotes.append(Messages.FitLog_CustomDataFields_Label);
          _currentActivity.customDataFields.forEach((key, value) -> {
             if (!tourNotes.toString().trim().isEmpty()) {
                //If there is already content in the notes fields, then we insert a new line
-               tourNotes.append(System.getProperty("line.separator")); //$NON-NLS-1$
+               tourNotes.append(UI.SYSTEM_NEW_LINE);
```
wolfgang-ch: You should use `UI.SYSTEM_NEW_LINE` only when necesary, e.g. export txt files in other cases `net.tourbook.common.UI.NEW_LINE` works for win & linux in the UI

wolfgang-ch:
FJBDev is now merged into master

PR #99, opened 2019-09-04
Adding an import preference setting to give the choice to set the body weight when saving an imported tour

FJBDev:


bundles/net.tourbook/src/net/tourbook/ui/views/rawData/RawDataView.java
```diff
@@ -4585,6 +4589,10 @@ private void restoreState() {
       final boolean isIgnoreInvalidFile = _state.getBoolean(STATE_IS_IGNORE_INVALID_FILE);
       _rawDataMgr.setState_IsIgnoreInvalidFile(isIgnoreInvalidFile);
 
+      // restore: set body weight status before the tours are imported
+      final boolean isSetBodyWeight = _state.getBoolean(STATE_IS_SET_BODY_WEIGHT);
```
wolfgang-ch: `_state.getBoolean(...` returns false when not yet saved

FJBDev: Fixed

bundles/net.tourbook/src/net/tourbook/messages.properties
```diff
@@ -1539,6 +1539,8 @@ PrefPage_Import_Checkbox_CreateTourIdWithTime_Tooltip = The unique tour ID can b
 PrefPage_Import_Checkbox_IgnoreInvalidFiles           = Ignore &invalid files when using the easy import
 PrefPage_Import_Checkbox_IgnoreInvalidFiles_Tooltip   = When using the Easy Import, invalid files will be saved (their absolute path and file names will be saved in the file \
                                                         "invalidfiles_to_ignore.txt") so that they are not considered files to be imported in the future.
+PrefPage_Import_Checkbox_SetBodyWeight                = Set the person's weight when saving an imported a tour
```
wolfgang-ch: "Set the person's weight when saving an imported a tour"

Is the "a" before "tour" correct? I would remove the "a"

When I translate the above text into german and then back into english (with google) then the text below is returned

"Set the weight of the person when saving an imported tour"

wolfgang-ch:
It would be better when the checkbox in the pref import page has a link to the person pref page from where the weight is comming from, similar to https://github.com/wolfgang-ch/mytourbook/blob/master/bundles/net.tourbook/src/net/tourbook/preferences/PrefPageComputedValues.java#L770-L776

bundles/net.tourbook/src/net/tourbook/messages.properties
FJBDev: Thanks for the catch. Fixed in the upcoming PR

FJBDev:
Fixed in upcoming PR

PR #101, opened 2019-09-05
Changing the weight value and label when the measurement system in MTB is changed

FJBDev:


wolfgang-ch:
A suprisingly new feature :-)

I discovered that in some places where the weight can be entered/saved, it is not converted metric <-> imperial.

FJBDev:
I had noted that myself but also some users noted it (see forum [here](https://sourceforge.net/p/mytourbook/discussion/622811/thread/6ba792c0ff/?limit=25#a722))

PR #102, opened 2019-09-06
Changes per code review and small bugfix

FJBDev:


wolfgang-ch:
In 

* net.tourbook.tour.DialogQuickEdit
* net.tourbook.ui.views.tourDataEditor.TourDataEditorView

the body weight is not converted in 3 of 4 cased, from UI (metric/imperial) to database (metric) and vice versa

wolfgang-ch:
net.tourbook.tour.TourInfoUI has also no metric/imperial conversion

`_lblBodyWeight.setText(_nf1.format(_tourData.getBodyWeight()));`

FJBDev:
Thanks for the catches, I fixed those in my PR

PR #106, opened 2019-09-13
Added new statistics to show the Hiking vs Running percentage

FJBDev:
Advised here  https://trainright.com/8-things-ultrarunners-do-improve-training/ by Jason
Koop, any ultra-runner should know how much hiking vs. running they will be
doing in your next race so they can get close to that ratio in training. 

image

image


wolfgang-ch:
Herea are a few issues which I've found

**TourData + other locations**

* Group names similar to `power_  weather_ ...`

```
cadenceZoneHikingTime	-> cadenceZone_HikingTime
cadenceZoneRunningTime	-> cadenceZone_RunningTime
```


**TourDatabase**

* Use `TOURBOOK_DB_VERSION = 40;` which is not yet published


**TourPerson**

* `cadenceZonesDelimiter` default value is 0, when cadence zone for all tours are computed and person is not resaved, then 0 will be the cadence zone delimiter



**TVITourBookItem**

* `result.getInt(startIndex + 22)` and `result.getInt(startIndex + 23)` should be put into a local var and not 2 times calculated, the same issue is in **TVITourBookYearSub**


**Other issues**

* It seems to be not final



* I don't understand 3 rpm values




* Maybe it should be possible to compute these cadence zone for selected tours, as it can be done in the tour book view for other tour values and the used cadence delimiter is stored in TourData

* Use `UI.EMPTY_STRING` instead of `""`

PR #109, opened 2019-09-18
Slow vs Fast cadence : code modifications per code review

FJBDev:
1) cadenceZoneHikingTime	-> cadenceZone_SlowTime
cadenceZoneRunningTime	-> cadenceZone_FastTime
2) Using TOURBOOK_DB_VERSION = 40;
3) cadenceZonesDelimiter default value is 70
4) Using local variables for result.getInt(startIndex + 22) and result.getInt(startIndex + 23) etc...
5) Improved the Cadence Zones tab in the People Preference page
6) Added the possibility to compute the cadence zones for selected tours
7) Use UI.EMPTY_STRING instead of ""


wolfgang-ch:
I found these issues

1. Cadence zone code is squeezed between hr zone code, cadence zone should be 80 and 80...hrZone should be move to 70...

source-numbering

2. The column title do not tell a lot

column-title

3. The actions in this sub-menu are sorted alphabetically and a mnemonic is missing, OK the SRTM action is also missing an mnemonic

action-sorting-mnemonic

4. 
* There is not much space between the label and the border, this can be solve with `GridLayoutFactory.swtDefaults()`
* There are hard breaks in the label text, see the text in the tab **Data Transfer** which is wrapping the text according to the width

swtfill-hard-break

5. In the early days of MT, I named the column with lowercases and many underscored, but since a while I changed the naming with less unterscores, e.g.

from
```
ColumnFactory_slow_vs_fast_cadence_percentages_header;
ColumnFactory_slow_vs_fast_cadence_percentages_label;
ColumnFactory_slow_vs_fast_cadence_percentages_tooltip;
```
to

```
ColumnFactory_SlowVsVast_CadencePercentages_Header;
ColumnFactory_SlowVsVast_CadencePercentages_Label;
ColumnFactory_SlowVsVast_CadencePercentages_Tooltip;
```

6. As I mentioned in my previous review, it would be helpfull to see which cadence delimiter is used

> and the used cadence delimiter is stored in TourData


### A lot of details but the UI should be consistently

FJBDev:
I have addressed all points [here](https://github.com/wolfgang-ch/mytourbook/pull/111)

PR #110, opened 2019-09-18
 Updating to Eclipse 4.13 (2019-09) 

FJBDev:
Eclipse 4.13 (2019-09) was just released today : https://www.eclipse.org/downloads/packages/release/2019-09


pom.xml
```diff
@@ -74,22 +74,22 @@
       
       <repository>
          <id>eclipse.platform.4.10</id>                                  <layout>p2</layout>
-         <url>jar:file:repositories/org.eclipse.platform-4.10.zip!/</url>
```
wolfgang-ch: The following 4 modified lines are within a comment and should not be modified.

The buildprocess for 19.10 worked with eclipse 4.13 repositoy

PR #111, opened 2019-09-18

FJBDev:
Addressed all the 6 bullet points mentioned [here](https://github.com/wolfgang-ch/mytourbook/pull/109)

For 6. As I mentioned in my previous review, it would be helpfull to see which cadence delimiter is used

I save the delimiter value in the TourData :
image


wolfgang-ch:
This morning, laying in bed, this thought came to me. 

Why is the cadence zone assigned to a person?

Sould it not be connected with an activity (tour type) or just to be general (in the first time) like the pref page/computed values/elevation gain or break time ?

Beside these thoughts, I found these issues.

1. Backslaph in the tooltip
backspace in text

2. Missing Cadence in the column label, what ist "Slow vs Fast"?
column-title

3. 
* mnemonic C is already used, d in Ca&dence could be used
* "Compute Time.." should be below "Compute Elevation..."
submenu

4.
* colon is not necessary
* mnemonic C do not work because the tab label has already a C
colon-mnemonic

wolfgang-ch:
This column is disabled
column-removed

FJBDev:
> 
> 
> This column is disabled
> column-removed

It was not used anywhere and was raising  a warning, thats why I commented it.

FJBDev:
This morning, laying in bed, this thought came to me.

Why is the cadence zone assigned to a person?

Sould it not be connected with an activity (tour type) or just to be general (in the first time) like the pref page/computed values/elevation gain or break time ?

It's a good question.
At first, I thought that cadence zones are different for each individual. But after your question, it's possible that it belongs to be general like the pref page/computed values/elevation gain or break time Do you have a preference ?

wolfgang-ch:
1. `defineColumn_Powertrain_PedalLeftRightBalance();` is used in master branch
https://github.com/wolfgang-ch/mytourbook/blob/master/bundles/net.tourbook/src/net/tourbook/ui/views/tourBook/TourBookView.java#L995

2. 
>Do you have a preference ?

I would move the cadence zone tab into _pref page/computed values_ and remove the cadence zone fields from the person

3. It would be helpful to see the `cadenceZones_DelimiterValue` in a new column in the tourbook view instead of selecting each tour to see this value, similar to this `net.tourbook.data.TourData.dpTolerance` field

PR #113, opened 2019-09-19

FJBDev:
Thank you for all the catches.

I have fixed what you mentioned in the previous code review :
1) defineColumn_Powertrain_PedalLeftRightBalance(); is used in master branch
Sorry, it was my mistake
2)I would move the cadence zone tab into pref page/computed values and remove the cadence zone fields from the person
3)It would be helpful to see the cadenceZones_DelimiterValue in a new column in the tourbook view instead of selecting each tour to see this value, similar to this net.tourbook.data.TourData.dpTolerance field
4)Backslaph in the tooltip
5)Missing Cadence in the column label, what ist "Slow vs Fast"?
6)mnemonic C is already used, d in Ca&dence could be used
"Compute Time.." should be below "Compute Elevation..."
7)colon is not necessary
mnemonic C do not work because the tab label has already a C


wolfgang-ch:
I found a few minor issues

1. Does it need to summarize the cadence zone delimiter value?
summarize-cadencezonedelimiter

2. Cadence zone is now in the `PrefPageComputedValues` and not in the people pref page, can be fixed easily with search/replace
pref-people-tag

3. Additional space
spacing

4. This text may be misleading _All existing..._ but the action is done only for the selected tours
info-message

FJBDev:
All minor issues fixed in latest Pull Request

PR #119, opened 2019-09-25
Using SQL statements to compute cadence zones times for multiple tours

FJBDev:


wolfgang-ch:
The log has no title line
no-log-title

FJBDev:
Fixed in my PR

PR #122, opened 2019-09-27
TCX Export: Adding the possibility to specify an activity type

FJBDev:


wolfgang-ch:
See my comments https://sourceforge.net/p/mytourbook/discussion/622811/thread/fb16b8ca26/#f935

bundles/net.tourbook.export/src/net/tourbook/export/DialogExportTour.java
```diff
@@ -188,7 +188,7 @@
    }
 
    private final static String[]     StravaActivityTypes = new String[] {
-         "Biking", "Running", "Hiking", "Walking", "Swimming"                                                                      //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$
+         "Biking", "Running", "Hiking", "Walking", "Swimming", "Other"                                                                      //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$
```
wolfgang-ch: Requires these texts string externalization ?

As the user can enter any value in this field, it would be helpful when previous entries could be selected, similar to the tour start/end texts
https://github.com/wolfgang-ch/mytourbook/blob/master/bundles/net.tourbook/src/net/tourbook/tour/DialogQuickEdit.java#L349-L356

FJBDev: Done, see my [PR](https://github.com/wolfgang-ch/mytourbook/pull/126)

PR #123, opened 2019-09-29
Per code review suggestions

FJBDev:


wolfgang-ch:
You started to change the wording from altitude to elevation, this is a long trip to do it everywhere :-)

The lowest pass which I've driven by car 2005 is in Florida

rock-reef-pass

FJBDev:
> 
> You started to change the wording from altitude to elevation, this is a long trip to do it everywhere :-)

Correct, it will be quite some work. But if you agree with it, here is my opinion : **Altitude** IS DIFFERENT than **Elevation Gain**
By definition : 
Altitude : the vertical elevation of an object above a surface (such as sea level or land) of a planet or natural satellite ([source](https://www.merriam-webster.com/dictionary/altitude))
Elevation gain :  cumulative elevation gain refers to the sum of every gain in elevation throughout an entire trip. ([Source](https://en.wikipedia.org/wiki/Cumulative_elevation_gain))

> 
> The lowest pass which I've driven by car 2005 is in Florida
> 
Wow, that's funny!

wolfgang-ch:
Yesterday, when I saw that your are using elevation then I found this short and simple description in wikipedia very helpful

> The term elevation is mainly used when referring to points on the Earth's surface, while altitude ... is used for points above the surface

https://en.wikipedia.org/wiki/Elevation

Here are a few points when you are changing the wording from altitude to elevation

- I would not change the text key only the text content
- Preserve mnemonics, when this is not possible check that the new mnemonic is unique which may be not possible in a few cases and depends which controls are enabled/disabled

I often use mnemonics e.g. in Eclipse to see the changed code with the head, press right mouse key + a + h https://bugs.eclipse.org/bugs/show_bug.cgi?id=475108, therefore a unique key is important (when possible)

FJBDev:
At first, I was only planning in changing altitude to "elevation gain" because they are two different concepts but I can also change to elevation when needed if you need me.
Ok, I will not change the text, ONLY the content.

PR #127, opened 2019-10-02
 Bugfix: Because of the extra space, the icons were not found and a red square was displayed instead

FJBDev:


FJBDev:
Before :
Before

After:
After

wolfgang-ch:
These space errors exist since more than 9 years :-)

wolfgang-ch:
As I remember, I saw the wrong cancel button in the pref dialog, however the cancel icon and the wrong square image are looking similar, both are red :-) 

I don't use or test the french version so I've not seen it in the 2D map.

PR #128, opened 2019-10-02
Changing only the text content and not the text keys 

FJBDev:


wolfgang-ch:
@FJBDev 
Should not the altitude be changed at other locations?

FJBDev:
Thanks, I will do those too.

Issue #125, opened 2019-09-30
endless loop when reading a CM4xxM DAT file


cyc1ingsir:
Hi,
the last file, I saved from my CicloSport CM414AM includes another very unlikely data constellation :-). Similar to this one:
https://github.com/wolfgang-ch/mytourbook/blob/53502fb8156f24da8449c75929090261210dd4b4/bundles/net.tourbook.device/src/net/tourbook/device/cm4xxm/CM4XXMDeviceReader.java#L536

The DD pointer discovered after the last tour in this case however, doesn't point to the initial AA record but to the one of the second tour.
I implemented a fix for this but the PR isn't quite ready yet.
Basically it is based on an implementation of an ancient project of mine:
https://gitlab.com/cyc1ingsir/openhac/blob/master/src/main/cm414tourlistfactory.cpp#L146

Even though the dangling DD pointer (in this case) points to a valid AA record by chance, 
the referenced AA record points to a different DD record. 
Hence, the DD record discovered is not a valid one.
A DD record may be validated by verifying the consistency of the backward as well as the forward reference:

``` java
        /*
	 * Verify that a DD record not only points to an AA record, 
	 * but the AA record pointed to, also references the same DD record.
	 * In case, it doesn't, a dangling / left over / invalid DD pointer was discovered.
	 * Even though, this is quite unlikely, it has happened before.
	 */
	private boolean verifyReferenceConsitency(final RandomAccessFile file, int offsetDDRecord) throws IOException {
		final byte[] buffer = new byte[5];
		String recordType = UI.EMPTY_STRING;
		
		file.seek(offsetDDRecord);
		file.read(buffer);
		recordType = new String(buffer, 2, 2);

		// make sure we read a DD record
		if (!recordType.equalsIgnoreCase("DD")) { //$NON-NLS-1$
			log.debug("No DD record found!");
			return false;
		}

		int offsetAARecord = DataUtil.readFileOffset(file, buffer);
		file.seek(offsetAARecord);
		file.read(buffer);
		recordType = new String(buffer, 2, 2);
		// make sure we read a DD record
		if (!recordType.equalsIgnoreCase("AA")) { //$NON-NLS-1$
			log.debug("No AA record found!");
			return false;
		}
		
		int forwardReference = DataUtil.readFileOffset(file, buffer);
		if (forwardReference != offsetDDRecord) {
			log.debug("DD -> AA -> DD record references not constistent!");
			return false;
		}
					
		return true;
	}
```

[...]

PR #130, opened 2019-10-03
Bugfix/125 endless loop cm4xxm

cyc1ingsir:
This PR provides a fix for #125 

I didn't manage to include the logging, therefore it's commented out. 
I wasn't able if there is a common ground on which logging framework to use (slf4j?) and how it's included. I found traces in the map25 subproject but I didn't manage to apply it in the modified class.


wolfgang-ch:
I've tested the import also with 2 other cm4xxm files (I have no more) and it seems to work, without your fix I can confirm the endless loop.

wolfgang-ch:
>I didn't manage to include the logging, therefore it's commented out.

When the easy import was implemented, a new logging "engine" was also created, `net.tourbook.tour.TourLogManager`

This is now mainly used for logging but maybe needs some improvements

PR #136, opened 2019-10-08
Multiple files Reimport fix

FJBDev:
Before :
image

After:
image


FJBDev:
While this Pull Request fixed an issue when reimporting multiple files, I have found an issue when reimporting multiple tours that were imported from the same unique file.
Per your comment [here](https://github.com/wolfgang-ch/mytourbook/blob/679dded1ae46fc7a1cad0ddb1e018195536deff7/bundles/net.tourbook/src/net/tourbook/importdata/RawDataManager.java#L1006), because you don't skip the tours that have not the same start time, then when I try to re-import, for example, 2 tours that were imported from 1 file that contained 10, then they can't be reimported because they are "not found" when in the reimporting code and the error message is as below:
When I uncomment //				continue;     then it works. I don't want to uncomment it because you did it for a good reason. Any ideas ??


02:05:37 |   | Reimport tour
-- | -- | --
02:05:40 |   | Tour "1/10/11, 7:20 PM" cannot be reimported from the file "C:\Users\frederic\Documents\Data files\FitLogs\ST Migration 3\2011.fitlogEx" because it doesn't contain the same number of time slices, current tour: 1720 - reimport file: 1
02:05:41 |   | Tour "1/11/11, 4:21 AM" cannot be reimported from the file "C:\Users\frederic\Documents\Data files\FitLogs\ST Migration 3\2011.fitlogEx" because it has a different tour ID, current tour: 20111114210 - reimport file: 201112419350
02:05:43 |   | Tour "1/11/11, 4:21 AM" cannot be reimported from the file "C:\Users\frederic\Documents\Data files\FitLogs\ST Migration 3\2011.fitlogEx" because it has a different tour ID, current tour: 20111114210 - reimport file: 201112419350
02:05:43 |   | Reimported in 6.756 s

FJBDev:
Or maybe if we should do that ??
```
if (timeDiff > 20000 && !fileName.toLowerCase().endsWith(".fit")){
continue;
}
```

FJBDev:
Good thing you found this issue now, thanks

wolfgang-ch:
I would be more happy with this 

``` 
if (timeDiff > 20000 && fileName.toLowerCase().endsWith(".fit") ==false) {
continue;
}
```

solution as it would not change the implemented logic

FJBDev:
Done. See my [PR](https://github.com/wolfgang-ch/mytourbook/pull/140)

PR #139, opened 2019-10-08
Escaped single quotes by a preceeding single quote.

FJBDev:
I was looking for that fix for a long time since the french language has a lot of single quotes '

Per the [documentation](https://help.eclipse.org/2019-09/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fosgi%2Futil%2FNLS.html)

> Text appearing within single quotes is treated as a literal. A single quote is escaped by a preceeding single quote. 

But that is annoying to have to adapt some of the french messages depending if they will be used in a `NLS.bind.`
@wolfgang-ch  any better ideas ?


wolfgang-ch:
`org.eclipse.osgi.util.NLS.bind` is (should be) only used when parameters `{0}, {1}, ...` are added otherwise it is not used, so you could identify the text strings.

FJBDev:
> 
> 
> `org.eclipse.osgi.util.NLS.bind` is (should be) only used when parameters `{0}, {1}, ...` are contained in the text, otherwise it is not used (and does not make sense to use it), so you can identify the text strings.

Ok, I suppose I only have to update the strings that are used within a `NLS.bind()`

FJBDev:
I have updated all the french string used in a `NLS.bind()` and containing an apostrophe.

Here is the RegEx I created in order to find them easily:
`^(?=.*\{[0-9]\})(?=.*\')(?!.*\'\').*$`

wolfgang-ch:
`^(?=.*\{[0-9]\})(?=.*\')(?!.*\'\').*$`

Something for advanced RegEx users :-)

Without your explanation I very likely do not know what is going on

PR #140, opened 2019-10-09
 Fixing the reimport for tours contained in files that have multiple tours while keeping it disabled for fit files

FJBDev:


wolfgang-ch:
@FJBDev 
Reimporting files causes this log enty

import-log

because the parameters for this code `String.format(NLS.bind(...` are wrongly implemented and this comination should be avoided, use either `String.format` or `NLS.bind`

string-format-nls-bind


It seems that this code modification was never tested.

FJBDev:
I tested this code but only in French and didn't realize at the time that it was impacting the other languages. Thanks for fixing

PR #147, opened 2019-10-13
Per code review

FJBDev:


FJBDev:
I found why Eclipse was changeing special characters to question marks. I fixed it by configuring file encoding to ISO-8859-1 as below : 
EclipseConfig

bundles/net.tourbook/src/net/tourbook/data/TourData.java
```diff
@@ -6673,6 +6673,13 @@ public void finalizeTour_SwimData(final TourData tourData, final List<SwimData>
       }
    }
 
+   /**
+    * @return the avgAltitudeChange
```
wolfgang-ch: @FJBDev When you reference a field in java doc, start by typing a hash character # then automatically a dropdown is opened with all fields and a link to the field can be inserted.

wolfgang-ch:
> I found why Eclipse was changing special characters to question marks. I fixed it by configuring file encoding to ISO-8859-1 as below :

The eclipse editor should be configured as in this config https://github.com/wolfgang-ch/mytourbook/blob/master/.editorconfig but I discovered, my current workspace was set to default, maybe I recreated the workspace and forgot it to set it correctly.

When setting my workspace to UTF-8 then your last modified version looks bad but TourData from 2019-03-05 is OK

utf8

As in .editorconfig is defined, I set the newline delimiter to Unix and when I open an existing file and modify it, then I also convert it to the Unix line delimiter when it's not yet set.

So we should stick to UTF-8 and Linux line delimiter.

I've also set git config accordingly
autocrlf

I just saved TourData again :-) but now with UTF-8 and correct ° 

This is now my workspace setup
workspace

bundles/net.tourbook/src/net/tourbook/data/TourData.java
FJBDev: Cool trick, I didn't know.
What I did was to make sure it was like the other get functions like [this one](https://github.com/wolfgang-ch/mytourbook/blob/5ce7c905fdc41333154698bdd903ea721994bd30/bundles/net.tourbook/src/net/tourbook/data/TourData.java#L6684)

`* @return the avgCadence`

FJBDev:
Ok, I will configure it as you described

bundles/net.tourbook/src/net/tourbook/data/TourData.java
FJBDev: In the future, I will put this : 
@return {@link #avgAltitudeChange}

PR #149, opened 2019-10-15
Translation issues

FJBDev:
WARNING! DON'T MERGE, THIS PULL REQUEST IS TO SHOW YOU THE PROBLEMS I HAVE WHEN WORKING WITH TRANSLATION FILES

@wolfgang-ch 
When I work on translation files (french usually), the indentation usually gets messed up and it's really annoying as I have to manually review every line and see which one was caused by Eclipse.
I have followed everything depicted [here](http://mytourbook.sourceforge.net/mytourbook/index.php/development/translation/install-resource-editor)
Here is my configuration in Eclipse 
image

image

Any idea what I am missing ?

### DIFF
--- a/bundles/net.tourbook/src/net/tourbook/messages_fr.properties
+++ b/bundles/net.tourbook/src/net/tourbook/messages_fr.properties
@@ -28,7 +28,7 @@ Action_PhotosAndTours_FilterPhotos_Tooltip         = Filtre parcours :\n\
                                                      Afficher seulement les parcours comprenant des photos
 Action_PhotosAndTours_SavePhotoInTour              = &Enregistrer les photos dans le parcours
 
-Action_TagCategory_Edit         = \u00C9diter la cat\u00E9gorie du Tag...
+Action_TagCategory_Edit         = \u00C9diter la cat\u00E9gorie du Tag..
 Action_TagCategory_EditCategory = \u00C9diter la cat\u00E9gorie...
 Action_TagCategory_Edit_Tooltip = \u00C9diter la cat\u00E9gorie du tag de tour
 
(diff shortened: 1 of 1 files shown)


bundles/net.tourbook/src/net/tourbook/messages_fr.properties
```diff
@@ -729,14 +731,14 @@ Dlg_AdjustAltitude_Radio_keep_start_altitude_tooltip  = Garder l'altitude du d\u
...
 Dlg_TourMarker_Button_PasteFromClipboard_Tooltip = Ins\u00E9rer le texte depuis le presse-papier et le convertir afin de le rendre plus lisible.\n\
-                                                   \n\
-                                                   Exemple: \n\
-                                                   \n\
-                                                   Presse Papier : \n\
-                                                   http://fr.wikipedia.org/wiki/L%27Alpe_d%27Huez\n\
-                                                   \n\
-                                                   Texte corrig\u00E9 : \n\
-                                                   http://fr.wikipedia.org/wiki/L'Alpe_d'Huez
+\n\
```
FJBDev: @wolfgang-ch Here is an example of what happens when I just save my french translations using with the Messages Editor. Do you know why this happens ?

wolfgang-ch: In my comment from Aug 23 https://github.com/wolfgang-ch/mytourbook/pull/95#pullrequestreview-279078143 I pointed out that long lines are formatted differently but the conent which is displayed to the user is the same

FJBDev: Ah true. I suppose if it's displayed to the user the same it's good. 
The only problem is that if i modify an english translation file and then you modify after me, then if I modify again, in the end it makes a lot of "noise".

wolfgang-ch: >in the end it makes a lot of "noise".

When you _Ignore White Space_, there is no noise, the yellow area is not marked as modified and this is the only difference I know, between the two messages editors

I modify other language files mainly, when adding new images that the translators do not have to "translate" it.

compare-noise

PR #156, opened 2019-11-07
 Adding the cadence zones time percentages in the weekly summary of the calendar

FJBDev:


bundles/net.tourbook/src/net/tourbook/ui/views/calendar/CalendarProfileManager.java
```diff
@@ -846,6 +850,53 @@ void setValueFormat(final ValueFormat valueFormat) {
...
+            FormatterID.CADENCE_ZONES_TIMES,
+            Messages.Calendar_Profile_Value_CadenceZones_TimePercentages,
+            GraphColorManager.PREF_GRAPH_CADENCE) {
+
+         @Override
+         String format(final CalendarTourData data, final ValueFormat valueFormat, final boolean isShowValueUnit) {
+
+            String cadenceZonesPercentages = UI.EMPTY_STRING;
+
+            final int totalCadenceTime = data.cadenceZone_SlowTime + data.cadenceZone_FastTime;
```
FJBDev: @wolfgang-ch Could you tell me where I could put this code in MTB in one unique place to avoid duplication as it is also [here ](https://github.com/wolfgang-ch/mytourbook/blob/c9fa3aa0e11a6cc491416ddbcb02975c2bc8ea3f/bundles/net.tourbook/src/net/tourbook/ui/views/tourBook/TVITourBookYearSub.java#L434) and [here ](https://github.com/wolfgang-ch/mytourbook/blob/c9fa3aa0e11a6cc491416ddbcb02975c2bc8ea3f/bundles/net.tourbook/src/net/tourbook/ui/views/tourBook/TVITourBookItem.java#L325)?

wolfgang-ch: This location has already cadence zone computation

https://github.com/wolfgang-ch/mytourbook/blob/master/bundles/net.tourbook/src/net/tourbook/tour/TourManager.java#L529

FJBDev: This location does something different. It computes the time (in seconds) spent in each zone. But the function I highlighted above generates a string with the percentage ratio as this : "33 - 67"

wolfgang-ch: I thought that you can put this code into TourManager.java and not use the code which I pointed to. I just pointed to this code because it is similar.

FJBDev: Oh, ok, i understand now. Will do !

PR #170, opened 2020-02-10
DialogEditTimeSlicesValues: Changes per code review

FJBDev:
1. Renamed the text keys from Dialog_Edit_* to Dialog_EditTimeslicesValues_*

2. When an item (altitude, cadence,...) is not selected (checked) then the corresponding fields are disabled.

3. Saving the dialog buttons and values to state for comfort to keep the state of the selected items/field values when the dialog is closed/reopened as this done also in other dialogs, e.g. export dialog.

4. Added the MT version number in the cheatsheet


wolfgang-ch:
@FJBDev I also did a few wording changes in the master branch. You may adjust the french translations. I'm planning to create MT 20.2 tomorrow morning (UTC +1h :-)

wolfgang-ch:
I wanted to publish MT 20.2 today, tested it on Ubuntu ->new dialog looks not very good

ubuntu-mt-20 2-time-slice-dialog

Using absolute values in the .indent() method is not a good idea

source-mt-20 2-time-slice-dialog

I started to adjust this dialog but it's not yet finished. 

I also already changed the SelectionListener -> SelectionAdapter -> less code

wolfgang-ch:
@FJBDev 
I forgot to mention you in the above comment, now you should receive a notification

FJBDev:
@wolfgang-ch 
Thanks, I will fix that.
Is there a way to fix the spinner widths? It seems that they have a different size depending if they have a decimal in it.

wolfgang-ch:
>Is there a way to fix the spinner widths?
as I remember my tests a long time ago, the width cannot be forced it depends on the numder of decimals as you also discovered

FJBDev:
It might be a Linux vs Windows behavior because when I try with Windows 10, the width is consistent

image

wolfgang-ch:
Yes, Windows looks better than Linux

I just commited my dialog adjustments, see also comments in https://github.com/wolfgang-ch/mytourbook/commit/65a63b35f00704a9b73af44d0903eda7b9181c70

I think this dialog can be dangerous to accidentally change the time slice values as this could not be done before. There is still the restore action after this dialog has updated the time slices.

mt-time-slice-editor

wolfgang-ch:
I just discovered a _hidden_ feature in Linux, when you right click on the + or - button of the spinner control, it's min/max value is entered in the value field

FJBDev:
> I think this dialog can be dangerous to accidentally change the time slice values as this could not be done before. There is still the restore action after this dialog has updated the time slices

It can be dangerous but the user can always restore the modifications

FJBDev:
> 
> 
> I just discovered a _hidden_ feature in Linux, when you right click on the + or - button of the spinner control, it's min/max value is entered in the value field

@wolfgang-ch  Actually, I have also discovered that if you press the mouse scroll wheel on the + or - buttons of the spinner control, it will increment/decrement the value by 10.
It's in Linux only and I can't find official doc for it yet.

PR #187, opened 2020-05-22
Dropbox Integration - Fixes per suggestions - Part 2

FJBDev:
- Dropdown menu should not be editable only selectable

image

</br>

- Dropbox is aligned as the other fields
image

</br>

- Displaying the chosen folder from the preferences in the import configuration
image

</br>

- We display the device column as the last column to see the other columns without horizontal scrolling

image
</br>

- Putting back the imports from 20.3 as I don't want to take risks breaking more things (cf. map25)

</br>

- In the import dialog, we display a link to the dropbox preference page if the dropbox token was not retrieved
image


wolfgang-ch:
👍

bundles/net.tourbook/src/net/tourbook/importdata/DialogEasyImportConfig.java
```diff
@@ -144,7 +144,7 @@
    private static final String          CSS_PX                            = "px";                                //$NON-NLS-1$
    private final static String[]        DeviceTypes                       = new String[] {
          Messages.Dialog_ImportConfig_Combo_Device_LocalDevice,
-         Messages.Dialog_ImportConfig_Combo_Device_Dropbox
+         "Dropbox"                                                                                               //$NON-NLS-1$
    };
```
wolfgang-ch, line 148: The `Dropbox` item is static, when the cloud plugin is not enabled, then Dropbox can currently still be selected. 
-> all items in DeviceType, except the local entry, must be provided from the available plugins, so DeviceType must be filled dynamically

In my first test of this PR I forgot to enable the cloud plugin and I just saw this issue.

wolfgang-ch:
When creating multiple easy import profiles then currently only a single Dropbox folder can be choosen for all profiles. This may be improved to select the dropbox folder in the easy import dialog instead of the pref page.

bundles/net.tourbook/src/net/tourbook/importdata/DialogEasyImportConfig.java
FJBDev: Fixed in this [PR](https://github.com/wolfgang-ch/mytourbook/pull/191)

FJBDev:
> When creating multiple easy import profiles then currently only a single Dropbox folder can be choosen for all profiles. This may be improved to select the dropbox folder in the easy import dialog instead of the pref page.

Fixed in this [PR](https://github.com/wolfgang-ch/mytourbook/pull/191)

PR #190, opened 2020-06-04
Map25: adding preference page for photolayer

telemaxx:
new preferencepage "Apperance" bellow map25 like in map2D
grafik

changing colors:
mapbookmark are blue like in toolbar(old: yellow)
photorating stars now yellow(old: black)


telemaxx:
grafik

wolfgang-ch:
> new preferencepage "Apperance" bellow map25 like in map2D
> grafik


Why are you not using the `2.5D Map Options` slideout or a new slideout to configure new features?
"Put the features closer to where it is used." this is the motto which I try to follow since a few years.


This was my tour on Wednesday :-)

telemaxx:
> Why are you not using the 2.5D Map Options slideout or a new slideout to configure new features?

yes, you're right, that is better.
and it seems to be easier to implement. i close this PR.

telemaxx:
what do you think of changing the colors?
map bookmarks in blue like in toolbar

83872229-da8aff80-a731-11ea-9cc9-93aeb2c6c8a3

wolfgang-ch:
A bookmark color should be customizable, it makes sense that the default color is blue like in the toolbar

telemaxx:
Button in GUI is implemented. But without any function yet.
grafik

now i will try to implement the functions of my "preference appearance page" branch.
maybe i can use "cherry picking" for that. i never used that...

PR #192, opened 2020-06-09
Adding the ability to select all the time slices in between the left and right sliders

FJBDev:
- Per requested in the feature [146](https://sourceforge.net/p/mytourbook/feature-requests/146/#0670/6fa0/aba8), the user can now choose to select automatically all the time slices between the left and right sliders with a new chart contextual menu

- Place the left slider and/or the right slider on the chart
- Right click on the left or right slider
- Click on "Select in-between time slices"
- If not already open, the Tour Data Editor View will open and the time slices between the left and right sliders will be selected.

**Before clicking**
image

**After clicking**
image


bundles/net.tourbook/src/net/tourbook/ui/views/tourDataEditor/TourDataEditorView.java
```diff
@@ -2468,6 +2471,27 @@ public void propertyChange(final PropertyChangeEvent event) {
...
+
+               if (_isSelectInBetweenTimeSlices) {
+                  final int minSelectedValue = Math.min(_lastSelectedChartInfo.leftSliderValuesIndex, _lastSelectedChartInfo.rightSliderValuesIndex);
+                  final int maxSelectedValue = Math.max(_lastSelectedChartInfo.leftSliderValuesIndex, _lastSelectedChartInfo.rightSliderValuesIndex);
+                  table.select(minSelectedValue, maxSelectedValue);
+               } else {
+                  table.select(selectionIndices[0]);
+               }
+               table.showSelection();
             }
```
wolfgang-ch, line 2495: Why have you implemented this feature with a pref store field ?

If I would have implemented all property changes with pref store fields there would be many of them. In this case, I would have create a new field in `SelectionChartInfo` with the state `isSelectInBetweenTimeSlices` and may be also in `SelectionChartXSliderPosition`

bundles/net.tourbook/src/net/tourbook/messages.properties
```diff
@@ -2788,84 +2786,87 @@ TourGeoFilter_Loader_LoadingError = Loading error:
...
+Tour_Action_MultiplyCaloriesBy1000_Title        = Multiply Calories Value
+Tour_Action_RetrieveWeatherData                 = Retrieve &Weather Data...
+Tour_Action_RunDyn_StanceTimeBalance_Tooltip    = Show/hide stance time balance
+Tour_Action_RunDyn_StanceTime_Tooltip           = Show/hide stance time 
+Tour_Action_RunDyn_StepLength_Tooltip           = Show/hide step length
+Tour_Action_RunDyn_VerticalOscillation_Tooltip  = Show/hide vertical oscillation
+Tour_Action_RunDyn_VerticalRatio_Tooltip        = Show/hide vertical ratio
+Tour_Action_Select_Inbetween_Timeslices         = Select in-between time slices
+Tour_Action_Select_Inbetween_Timeslices_Tooltip = If checked, selects all the time slices in the Tour Editor View between the left and right sliders. Otherwise, selects only the current slider's \
+                                                  time slice.
```
wolfgang-ch, line 2833: "If checked, selects all the time slices in the Tour Editor View between the left and right sliders. Otherwise, selects only the current slider's time slice."

"If checked, selects all the time slices in the Tour Editor View between the left and right sliders. Otherwise **it** selects only the current slider's time slice."

Wording? You are the english/french man :-)

bundles/net.tourbook/src/net/tourbook/ui/tourChart/SlideoutTourChartOptions.java
```diff
@@ -44,316 +44,350 @@
...
+            : X_AXIS_START_TIME.START_WITH_0;
 
-		_chkShowStartTimeOnXAxis.setEnabled(canShowTimeOnXAxis);
-		_chkShowStartTimeOnXAxis.setSelection(tcc.xAxisTime == X_AXIS_START_TIME.TOUR_START_TIME);
+      /*
+       * Update pref store
+       */
+      _prefStore.setValue(ITourbookPreferences.GRAPH_IS_BREAKTIME_VALUES_VISIBLE, isShowBreaktimeValues);
+      _prefStore.setValue(ITourbookPreferences.GRAPH_IS_SRTM_VISIBLE, isSrtmDataVisible);
+      _prefStore.setValue(ITourbookPreferences.GRAPH_X_AXIS_STARTTIME, isTourStartTime);
```
wolfgang-ch, line 379: In the beginning of MT, I also used pref store fields but in this tour chart options dialog you can see the mixture of different state fields.

bundles/net.tourbook/src/net/tourbook/ui/views/tourDataEditor/TourDataEditorView.java
FJBDev: Because I saw that [VALUE_POINT_TOOL_TIP_IS_VISIBLE ](https://github.com/wolfgang-ch/mytourbook/blob/d2cdac9cc46d89e6948e4fe2c90ab6a19f253998/bundles/net.tourbook/src/net/tourbook/preferences/ITourbookPreferences.java#L541) was implemented in the pref store too

I will change based on your recommendation

bundles/net.tourbook/src/net/tourbook/messages.properties
FJBDev: Will fix!

bundles/net.tourbook/src/net/tourbook/ui/tourChart/SlideoutTourChartOptions.java
FJBDev: Ok, good to know

bundles/net.tourbook/src/net/tourbook/ui/views/tourDataEditor/TourDataEditorView.java
wolfgang-ch: Sometimes it is OK when using the pref store field but in this case it can be avoided.

FJBDev: After thinking about it, and to offer to the user a better experience, I have changed the UI for this feature.

wolfgang-ch:
Have you tested this case:

* Open tour editor
* Select a tour (in any directory view)
* Edit tour in tour editor but don't save it
* Select another tour (in any directory view)
* Select your new action

When working with the tour editor it is always critical, if it is edited but not yet saved tours.

FJBDev:
> 
> 
> Have you tested this case:
> 
>     * Open tour editor
> 
>     * Select a tour (in any directory view)
> 
>     * Edit tour in tour editor but don't save it
> 
>     * Select another tour (in any directory view)
> 
>     * Select your new action
> 
> 
> When working with the tour editor it is always critical, if it is edited but not yet saved tours.

No, i didn't think about that case, good idea, i will test and modify if necessary

FJBDev:
> > When working with the tour editor it is always critical, if it is edited but not yet saved tours.

Looking at it closer, it seems that I should disable the new action in this case as my tests seem to show that if another tour is modified, the time slices cannot be changed to another tour and will stay on the current on modified.

is that a correct assumption ?

FJBDev:
See my latest update where I prevent from selecting the time slices if a tour is not saved.

bundles/net.tourbook/src/net/tourbook/ui/tourChart/action/ActionSelectInBetweenTimeSlices.java
```diff
@@ -0,0 +1,65 @@
...
+
+      //Opens the TimeSlice tab
+      tourDataEditorView.selectTimeSlicesTab();
+
+      tourDataEditorView.setRowEditModeEnabled(true);
+
+      tourDataEditorView.selectTimeSlice_InViewer(selectionChartInfo.leftSliderValuesIndex, selectionChartInfo.rightSliderValuesIndex);
+
+      //We recreate the viewer to take into account the new selection
+      tourDataEditorView.recreateViewer();
```
wolfgang-ch, line 63: Why not putting all tour editor methods directly into one method in the tour editor, e.g. `actionSelectInBetweenTimeSlices()`?

This also have the advantage to not set some methods public.

Why are you using `tourDataEditorView.recreateViewer();` this is a very expensive methods and have a signifant delay when you have >10'000 time slices or a >3h tour?

wolfgang-ch: Instead of an action to select time slices in the tour editor, this feature may also be a flag that the movement of the sliders is selecting immediately the time slices, either one (how it currently works) or all between.

wolfgang-ch: Instead of a checkbox action, a checkbox in the chart options would be a better solution for my previous comment

chart-option

FJBDev: >Instead of a checkbox action, a checkbox in the chart options would be a better solution for my previous comment

I had that originally but I didn't like it because it might use unnecessary extra CPU cycles if the user keeps it checked and moves the sliders but doesn't need to select the time slices.
Hence why I changed to an action that a user clicks when it's really needed.

Let me know what you prefer.

FJBDev: > Why are you using tourDataEditorView.recreateViewer(); this is a very expensive methods and have a signifant delay when you have >10'000 time slices or a >3h tour?

Because otherwise the new selected time slices are not selected in the viewer. So this is the only solution I found to refresh the viewer and see the selected new time slices.

wolfgang-ch: > I had that originally but after thinking about it I didn't like it because it might use unnecessary extra CPU cycles if the user keeps it checked and moves the sliders but doesn't need to select the time slices.
> Hence why I changed to an action that a user clicks when it's really needed.
> 
> Let me know what you prefer.

In MT there are some cases where this situation occures but it could be solved relative easily by waiting for a few ms and do the action when the user has stopped, e.g. you can find a similar case here

https://github.com/wolfgang-ch/mytourbook/blob/150ab2b37685edb7468d2b55c85f4e0bc1a35bab/bundles/net.tourbook/src/net/tourbook/map2/view/Map2View.java#L2064-L2100

When searching MT for `timerExec()` there will be some findings, however not only for this situation. The most prominent (but more complicated) case is the drawing of the tour chart which is also delayed and do not happen when resizing the chart window.

wolfgang-ch: > > Why are you using tourDataEditorView.recreateViewer(); this is a very expensive methods and have a signifant delay when you have >10'000 time slices or a >3h tour?
> 
> Because otherwise the new selected time slices are not selected in the viewer. So this is the only solution I found to refresh the viewer and see the selected new time slices.
> Ideas ?

I think it should be possible and will try to solve it when you've fixed the above issue.

FJBDev: > 
> 
> > > Why are you using tourDataEditorView.recreateViewer(); this is a very expensive methods and have a signifant delay when you have >10'000 time slices or a >3h tour?
> > 
> > 
> > Because otherwise the new selected time slices are not selected in the viewer. So this is the only solution I found to refresh the viewer and see the selected new time slices.
> > Ideas ?
> 
> I think it should be possible and will try to solve it when you've fixed the above issue.

I am confused : If I implement the "_above issue_" like you say (i.e. : _a checkbox in the chart options would be a better solution for my previous comment_) then I don't need to call this `tourDataEditorView.recreateViewer()` as when the user moves the sliders, the time slices will be selected in the background but the viewer will not need to be recreated.
Does that make sense ?

wolfgang-ch: >then I don't need to call this `tourDataEditorView.recreateViewer()` as when the user moves the sliders, the time slices will be selected in the background but the viewer will not need to be recreated.

When you implement this feature with the `timerExec()` solution then a recreation of the viewer is not necessary

FJBDev: Addressed in this [PR](https://github.com/wolfgang-ch/mytourbook/pull/192)

PR #197, opened 2020-06-27
Upstream/bug115

rtdog:
I'd like to submit for a minor bug fix related to bug 115.  It's a pretty simple change and code was modified to match technique used on other text fields StartLocation and EndLocation  I'm new to MTB and a long time ST3 user that's looking for a new system to manage my fitness logs.

### DIFF
--- a/bundles/net.tourbook/src/net/tourbook/ui/views/tourDataEditor/TourDataEditorView.java
+++ b/bundles/net.tourbook/src/net/tourbook/ui/views/tourDataEditor/TourDataEditorView.java
@@ -377,8 +377,7 @@ public class TourDataEditorView extends ViewPart implements ISaveablePart, ISave
     * contains the tour id from the last selection event
     */
    private Long                               _selectionTourId;
-   //
-   private KeyAdapter                         _keyListener;
+   //   
    private ModifyListener                     _modifyListener;
    private ModifyListener                     _modifyListener_Temperature;
    private MouseWheelListener                 _mouseWheelListener;
@@ -2870,13 +2869,7 @@ public class TourDataEditorView extends ViewPart implements ISaveablePart, ISave
             _isTemperatureManuallyModified = true;
             setTourDirty();
          }
-      };
-      _keyListener = new KeyAdapter() {
-         @Override
-         public void keyReleased(final KeyEvent e) {
-            onModifyContent();
-         }
-      };
+      };      
 
       /*
        * listener for recording/driving/paused time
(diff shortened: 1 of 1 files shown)


wolfgang-ch:
Hi @rtdog 

Welcome to the MT contributors, it looks like that you have some development skills :-)

To test your PR, I want to merge it but not with the `main` branch, therefore this `Common_PullRequests_Branch` branch is available which should be in sync with `main`.

Your PR is not large but I want to test the replaced `keylistener`, it's a long time ago it was implemented and can't remember why I've choosen this solution instead of the `verifylistener`

rtdog:
Thanks and pleased to meet you @wolfgang-ch .  To be honest I was so excited to discover MT that I implemented a number of features, but then only submitted this particular bug fix.  I'm still learning GitHub and and had to rework my branching strategy and untangle the various code changes I made to get just this bug fix, so there is some risk that I missed something in that untangle.  

Let me know how you'd like to get pull requests, if they always go to Common_PullRequests_Branch then I'm happy to do that.  I'm also curious how developers test with end users prior to submitting pull requests.  If end users aren't familiar with development tools, etc., then I'm not sure how they can test without either a binary distribution or some type of pluggable test framework.  I know Eclipse has that "Install New Software" feature which might help, what is the historic practice on end user testing been?

wolfgang-ch:
> Let me know how you'd like to get pull requests, if they always go to Common_PullRequests_Branch then I'm happy to do that. 

The branch `Common_PullRequests_Branch` is the default branch for new contributors but when they are committing regularly then they can have their own branch.

>I'm also curious how developers test with end users prior to submitting pull requests. If end users aren't familiar with development tools, etc., then I'm not sure how they can test without either a binary distribution or some type of pluggable test framework. I know Eclipse has that "Install New Software" feature which might help, what is the historic practice on end user testing been?

There is no end users test, except in a few cases I published also a test version which is difficult because the end user should not use it's "production" tour data otherwise the database version is locked and cannot be changed in the released version. 

In the last years, when a new version was ready, I do the german translation and create a test version for the translators and give them 2 weeks for doing the translation. During this 2 weeks I also use the test version with my production tour data and also update the website.

The built process of the product is a one click operation (when all is setup correctly), you can find more info in https://github.com/wolfgang-ch/mytourbook/blob/main/info/_HELP-create-build.txt

wolfgang-ch:
@rtdog 
I'm planning to create a new version by the end of the month (20.7) and would like to include your PR. But first I want to test it, so can you setup this PR to the `Common_PullRequests_Branch` because I do not merge it with this github tool into the `main` branch

rtdog:
@wolfgang-ch , yes, I'll be happy to issue the pull request into that Common_PullRequests_Branch .  It might take me a couple of days to get that done, but I'll get on it.  Thanks!

rtdog:
Ok @wolfgang-ch , I think I got it to the Common_PullRequests_Branch.  Should this one be closed?

wolfgang-ch:
fixed in other PR

PR #198, opened 2020-06-29
Select all the time slices in between the left and right sliders :  Instead of a checkbox action, a checkbox in the chart options would be a better solution

FJBDev:
Per this [code review suggestion](https://github.com/wolfgang-ch/mytourbook/pull/192#discussion_r442684916)

Not sure that I understood fully what you were suggesting, so feel free to correct me if I implemented something different.
Also, see my 2 comments where I got stuck and made me question if I understood and implemented your suggestion incorrectly.

### DIFF
--- a/bundles/net.tourbook/src/net/tourbook/ui/tourChart/TourChartView.java
+++ b/bundles/net.tourbook/src/net/tourbook/ui/tourChart/TourChartView.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (C) 2005, 2019 Wolfgang Schramm and Contributors
+ * Copyright (C) 2005, 2020 Wolfgang Schramm and Contributors
  *
  * This program is free software; you can redistribute it and/or modify it under
  * the terms of the GNU General Public License as published by the Free Software
@@ -592,7 +592,9 @@ public class TourChartView extends ViewPart implements ITourChartViewer, IPhotoE
                            new SelectionChartXSliderPosition(//
                                  _tourChart,
                                  tourDataSelection.getLeftSliderValueIndex(),
-                                 tourDataSelection.getRightSliderValueIndex()));
+                                 tourDataSelection.getRightSliderValueIndex(),
+                                 _prefStore.getDefaultBoolean(
+                                       ITourbookPreferences.TOGGLE_STATE_SELECT_INBETWEEN_TIME_SLICES)));
                   }
                }
             }
@@ -785,7 +787,9 @@ public class TourChartView extends ViewPart implements ITourChartViewer, IPhotoE
                _tourChart,
                SelectionChartXSliderPosition.IGNORE_SLIDER_POSITION,
                leftSliderValueIndex,
-               rightSliderValueIndex);
+               rightSliderValueIndex,
+               _prefStore.getDefaultBoolean(
+                     ITourbookPreferences.TOGGLE_STATE_SELECT_INBETWEEN_TIME_SLICES));
 
          xSliderPosition.setCenterSliderPosition(true);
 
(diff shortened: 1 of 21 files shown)


wolfgang-ch:
Because I've seen you have not yet implemented the `timerExec()` version, I'm waiting for it but I have already seen that this tooltip needs line wrapping

tooltip

bundles/net.tourbook/src/net/tourbook/ui/tourChart/TourChartView.java
```diff
@@ -592,7 +592,9 @@ private void onSelectionChanged(final ISelection selection) {
                            new SelectionChartXSliderPosition(//
                                  _tourChart,
                                  tourDataSelection.getLeftSliderValueIndex(),
-                                 tourDataSelection.getRightSliderValueIndex()));
+                                 tourDataSelection.getRightSliderValueIndex(),
+                                 _prefStore.getDefaultBoolean(
+                                       ITourbookPreferences.TOGGLE_STATE_SELECT_INBETWEEN_TIME_SLICES)));
```
wolfgang-ch, line 597: This code is used in many places, would it not be easier to put it in the `SelectionChartXSliderPosition` constructor or read this value in the tour editor where the selection is performed?

FJBDev: >>would it not be easier to put it in the SelectionChartXSliderPosition
I can't do that as otherwise it creates a circular dependencies
image


>>or read this value in the tour editor where the selection is performed?

That's what I did in my first [original PR](https://github.com/wolfgang-ch/mytourbook/pull/192) but you told me not to

image

FJBDev:
> 
> 
> Because I've seen you have not yet implemented the `timerExec()` version, I'm waiting for it but I have already seen that this tooltip needs line wrapping
> 
> tooltip

Do I still need to impement the `timerExec()` version since I don't have the checkbox action, but instead a checkbox in the chart options ??

Otherwise I have fixed the line wrapping in the tooltip

wolfgang-ch:
>Do I still need to impement the timerExec() version since I don't have the checkbox action, but instead a checkbox in the chart options ??

When only a few (<100) time slices are selected then it is OK without timerExec() but then there are more e.g. thousands then the UI is far from being smoothly.

FJBDev:
> When only a few (<100) time slices are selected then it is OK without timerExec() but when there are more e.g. thousands then the UI is far from being smoothly.

Ok, I will do it then

wolfgang-ch:
>>or read this value in the tour editor where the selection is performed?

>That's what I did in my first original PR but you told me not to

When a setting is used in different locations, then it's the easiest way to use a pref field which in this case, is the preferable way. But keep in mind that a pref field value change will fire an event to all pref listeners, this is the reason why I try to avoid pref fields when possible.

I would prefer to read the pref field in the tour editor and remove the additonal parameter `isSelectInBetweenTimeSlices` in `SelectionChartXSliderPosition`

When not implementing a feature by myself and review other work, then it's sometimes difficult to see the obstacles.

FJBDev:
> When a setting is used in different locations, then it's the easiest way to use a pref field which in this case, is the preferable way. But keep in mind that a pref field value change will fire an event to all pref listeners, this is the reason why I try to avoid pref fields when possible.
> 
> I would prefer to read the pref field in the tour editor and remove the additonal parameter `isSelectInBetweenTimeSlices` in `SelectionChartXSliderPosition`
> 
> When not implementing a feature by myself and review other work, then it's sometimes difficult to see the obstacles.

For sure, i understand it's difficult. Because I am still new to MT, I always assume you know better :-)
I will implement all that.

bundles/net.tourbook/src/net/tourbook/ui/tourChart/TourChartView.java
FJBDev: Addressed in this [PR](https://github.com/wolfgang-ch/mytourbook/pull/192)

PR #200, opened 2020-07-12
Update TourDataEditorView.java - Bug 115

rtdog:
Modified Title field to match technique used on other text fields StartLocation and EndLocation.  Added selectionListener for FTP and adjusted Wind Speed field so that conversion from Imperial to Metric units works a little better.


rtdog:
Sorry, this is in the wrong place...

PR #201, opened 2020-07-12
Update TourDataEditorView.java - Bug 115

rtdog:
Modified Title field to match technique used on other text fields StartLocation and EndLocation. Added selectionListener for FTP and adjusted Wind Speed field so that conversion from Imperial to Metric units works a little better.


wolfgang-ch:
Thanks for fixing this issue.

The next time when you create a PR for `Common_PullRequests_Branch`, first you should merge branch `main` into `Common_PullRequests_Branch` otherwise there can be many differences.

wolfgang-ch:
bug [#115](https://sourceforge.net/p/mytourbook/bugs/115/)

PR #204, opened 2020-07-16
Feature 178 - Improved imperial unit support

rtdog:
Feature - 178. The new start wizard now prompts for units before the first person which allows a user to select imperial units prior to configuring person. I also added support for feet and inches for imperial unit users on the Person configuration instead of meters.

### DIFF
--- a/bundles/net.tourbook.common/src/net/tourbook/common/Messages.java
+++ b/bundles/net.tourbook.common/src/net/tourbook/common/Messages.java
@@ -36,6 +36,7 @@ public class Messages extends NLS {
    public static String        Action_ColumnManager_Column_Info;
    public static String        Action_ColumnManager_ColumnActions_Info;
    public static String        Action_ColumnManager_FreezeCurrentColumn;
+   public static String        Action_ColumnManager_FreezeCurrentColumn_Tooltip;
    public static String        Action_ColumnManager_HideCurrentColumn;
    public static String        Action_ColumnManager_Profile_Info;
    public static String        Action_ColumnManager_ShowAllColumns;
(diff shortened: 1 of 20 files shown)


wolfgang-ch:
Feature [178](https://sourceforge.net/p/mytourbook/feature-requests/178/)

wolfgang-ch:
Almost perfect :-)

Because you changed the dialog sequence, this should also be reflected in this dialog

mt-1st-start-dialog 

by swaping `User` and `Measurement System`

Normally, when you change a translatable text, you **MUST** not modify the existing texts because text translators have no possibility to easily identify these changes. Therefor you must delete the text for **ALL** languages and create new text keys with another key name.

In this special case, where existing text is not changed only moved, you can do it for all languages in the existing text key.

Because of the many changes in a release, they are tracked in this file
https://github.com/wolfgang-ch/mytourbook/blob/main/info/release-notes/20.month.0-readme.txt

Can you add a short description of your feature into this file, also your _(name)_. This file will be published with the new release.

@rtdog @FJBDev @telemaxx 
All written texts/images in your PR's are more or less useless when they cannot be found easily.
My suggestion is to add links to your PR also into the release notes which will be the base for the texts on the website.

FJBDev:
> 
> 
> Feature - 178. The new start wizard now prompts for units before the first person which allows a user to select imperial units prior to configuring person. I also added support for feet and inches for imperial unit users on the Person configuration instead of meters.

Another thing that could be improved/modified would be the "km" icon in the Tour Chart. I find it confusing because the user could think that clicking on it will display kilometers but it will actually display the distance (in miles or km)
Maybe it could be changed to "km/mi" ?

image

Thoughts ?

wolfgang-ch:
Instead of "km" icon we can display a "mi" icon depending on the current measurement system

FJBDev:
> Instead of "km" icon we can display a "mi" icon depending on the current measurement system

Good idea, I have done so in my current [PR](https://github.com/wolfgang-ch/mytourbook/pull/208)

rtdog:
@wolfgang-ch , I updated the readme per your request in PR https://github.com/wolfgang-ch/mytourbook/pull/210 .  I see that you already updated the language files, thanks for doing that.

PR #216, opened 2020-08-03
Feature 170 (with good formatting)

rtdog:


wolfgang-ch:
@rtdog 

First it looked very good with a few minor issues (which I would have fixed for MT 20.8) but then I found an issue which stopped me to use this feature in MT 20.8.

This screenshot shows the cadende feature with Czech language

mt-cadence

When I reviewed the code then I discovered that the code saves `rpm` or `spm` but this do not work with Czech language, the default value is not be set in this dialog.

I think you should use an enum for the saved state, enums are already used in a few cases for state values.

Not all of my tour types have rpm or spm values, e.g. `Jet Aircraft` so a third cadence value should also be selectable, e.g. `- No Cadence -`

## Minor issues
* Cadence mnemonic c in launcher config is already used 
* Method name should start with a lowercase character: `net.tourbook.importdata.RawDataManager.DefaultCadence()`
* You modified and reverted `net.tourbook.preferences.ITourbookPreferences.IMPORT_IS_NEW_UI` but without `//$NON-NLS-1$`

rtdog:
Thanks for finding this @wolfgang-ch .  I'll get these fixed up for the next release.  The Czech language appears to be using a non-ASCII character which likely doesn't save into the settings file well.  An enum is a better idea.  

The DB is using a float cadenceMultiplier for rpm/spm.  I guess we could use 0.0f for the "no cadence" support above, otherwise this feature gets more complicated.  The Tour Editor View also uses radio buttons for the spm/rpm selection and there is currently no option for no value.  If you opened your Jet Aircraft tour right now in the Tour Editor I would expect it to default to rpm and upon save then it would have rpm (1.0f) stored as the cadenceMultiplier.  

This feature hasn't yet modified the Tour Editor screen, but will need to in order to add this new value.  Furthermore the rick-click "Adjust Tour Values" menu will need to have the additional selection available.  

It also appears the "Reimport Tour" will set the cadence value back to rpm, although it has likely always done this.  I'm not sure how to better handle a tour reimport.

wolfgang-ch:
0.0f for `No Cadence` is OK.

CadenceMultiplier default value in the DB is 1.0f

I'm aware that a new `No Cadence` state needs adjustments in some places in the app, a good chance to learn more about MT code :-)

For the `Tour Reimport` we should keep it as is, this tool is used not very often. Currently the reimport do also not respect other settings from the easy import configuration (as I remember).

bundles/net.tourbook/src/net/tourbook/preferences/PrefPageImport.java
```diff
@@ -42,6 +44,7 @@
 
    public static final String    ID          = "net.tourbook.preferences.PrefPageImport"; //$NON-NLS-1$
 
+   private final IPreferenceStore _prefStore  = TourbookPlugin.getPrefStore();
```
wolfgang-ch, line 47: This code is not autoformatted, here it's just one space but it looks like you have not activate the autosave feature in Eclipse http://mytourbook.sourceforge.net/mytourbook/index.php/development/dev-ide/code-style which will autoformat the code when it is saved.

This field is also not used but not used fields are not automatically removed. It could be set to remove unused fields (as I remember) but this would be too dangerous especially when you have not finished the code.
mt-not-used

PR #221, opened 2020-08-14
Cadence feature supported as enum

rtdog:
Changed cadence to enum and added "none" cadence type.


rtdog:
Cadence is not combo box, no longer radio buttons and  have three values: none, rpm, spm:

image

bundles/net.tourbook/src/net/tourbook/tour/Cadence.java
```diff
@@ -0,0 +1,55 @@
...
+   public static Cadence getByValue(final int value) {
+
+      for (final Cadence type : Cadence.values()) {
+         if (value == type.value) {
+            return type;
+         }
+      }
+
+      return Cadence.INVALID;
+   }
```
wolfgang-ch, line 41: The cadence values are saved as integer values which makes it not very easy when viewing the xml file. 

There are already some helper functions to read/write enum values as text:

net.tourbook.common.util.Util.getStateEnum(...)
net.tourbook.common.util.Util.setStateEnum(...)

rtdog:
Ok, I switched to text-based storage for enum instead of ints...

wolfgang-ch:
I've not yet tested the functionality, now I merged it that I can compare your branch with main branch in Eclipse and found these issues:

## 1. Cadence value is saved as string or integer

multiple-types

## 2. Code could be simplified
by using getXmlEnum() or setXmlEnum()

code-simplified

## 3. Multiple methods for returning the same value

multiple-methods

## 4. Reformatted code

reformatted-code

reformatted-code-2

## 5. No externalized strings in a few locations

externalize-strings


## 6. Not used code could be removed

not-used-code


## 7. Common text's are prefixed with `App_`

In this case `App_Unit_Cadence...` would be suitable

common-texts


## 8. 20.8 readme file is deleted
readme-is-deleted

Before you create a PR you can easily compare 2 branches and see all you changes at once.

git-compare


## 9. Rename `net.tourbook.tour.Cadence`
The word `Cadence` is used in many places with different meanings, e.g. cadence unit, cadence value or cadence label, so it would be easier to read the code when it's renamed into e.g. `net.tourbook.tour.CadenceMultiplier`. This can be done easily for Java with the refactoring tool.

rtdog:
Thanks for the feedback here @wolfgang-ch .  I have struggled to get the git compare tools to work in Eclipse despite some effort.  I tried a different diff technique this time around to see if I can avoid some of these mistakes.  You can see [PR#224](https://github.com/wolfgang-ch/mytourbook/pull/224) for this attempt.

wolfgang-ch:
When I compare 2 commits, then I do it with Eclipse however to sove git conflicts with Eclipse is not very good or I've not yet understood it, I solve conflicts with http://gitextensions.github.io/

bundles/net.tourbook/src/net/tourbook/importdata/EasyImportManager.java
```diff
@@ -1278,6 +1278,6 @@ private void setTourType(final TourData tourData, final ImportLauncher importLau
             String.format(//
                   LOG_EASY_IMPORT_003_TOUR_TYPE_ITEM,
                   tourData.getTourStartTime().format(TimeTools.Formatter_DateTime_S),
-                  tourTypeName + " (" + tourTypeCadence + ")"));
+                  tourTypeName + " (" + tourTypeCadence.getNlsLabel() + ")"));
```
wolfgang-ch, line 1281: I do this comment in the old PR because I cannot do it in the new PR.

You should run the Eclipse tool `Externalize Strings` that all of your new strings are externalized or ignored.

Issue #229, opened 2020-09-18
Paused time has wrong values


wolfgang-ch:
> 
> 
> I've several tours with wrong paused time, these are mainly tours which I recorded when driving by car but I also found 1 normal bicycle tour with wrong paused time value. These tours are all recorded with an Edge 1030 -> FIT format.
> 
> ![mt-stat-values](https://user-images.githubusercontent.com/1283445/93472071-7cef7d00-f8f4-11ea-97d9-eda9da3d6bd1.png)

I can't reproduce the issue so I'm not sure I can troubleshoot....unless you can share one or more FIT files ?

_Originally posted by @FJBDev in https://github.com/wolfgang-ch/mytourbook/pull/227#issuecomment-694406240_

wolfgang-ch:
I will investigate this, I've found more extreme examples. Your fix for negative paused times are not included in this test.

![mt-elapsed-vs-paused](https://user-images.githubusercontent.com/1283445/93598795-82fd6080-f9bd-11ea-9658-fd89be8965f9.png)

![mt-elapsed-vs-paused-2](https://user-images.githubusercontent.com/1283445/93598797-8395f700-f9bd-11ea-8078-a5bfced0e69a.png)

wolfgang-ch:
When dumping the paused start/end time, then there are some items where the start time is before the previous end time

![mt-next-start-before-end](https://user-images.githubusercontent.com/1283445/93611588-c7ddc300-f9ce-11ea-88e7-b0138940bf1c.png)

wolfgang-ch:
Tour Book view shows the computed diff: elapsed-recorded=paused

The statistic view shows the saved paused time which is computed with

https://github.com/wolfgang-ch/mytourbook/blob/64d8b860559b80e7f4cac8c761865b592bfa2d15/bundles/net.tourbook/src/net/tourbook/data/TourData.java#L8326-L8339

@FJBDev We have a big problem with the paused time :-)

How could it be solved when a paused start time is before the previous end time?
What value is correct?

FJBDev:
> 
> 
> I will investigate this, I've found more extreme examples. Your fix for negative paused times are not included in this test.
> 
> ![mt-elapsed-vs-paused](https://user-images.githubusercontent.com/1283445/93598795-82fd6080-f9bd-11ea-9658-fd89be8965f9.png)
> 
[...]

PR #233, opened 2020-09-21
Support for recorded and paused time Fixes Part2

FJBDev:
Code fixes for [this issue](https://github.com/wolfgang-ch/mytourbook/issues/229#issuecomment-695181871) #229

General code improvements


wolfgang-ch:
After the paused times are now working with my .fit data, it's a very cool feature :-)

Short pauses are now also visible in the timeline, without the displayed paused value, a pause could be hidden.

A problem is still to overlapped paused time which I already mentioned. When this should be solved, the tour marker labels should also be considered in the stacked labels which can be more complex because of the different tour marker label positions (vertical, horizontal)

mt-pauses

wolfgang-ch:
I did some modifications in the pref page UI, see my commit comments in https://github.com/wolfgang-ch/mytourbook/commit/dc369ab7510f6e1dc40a9840e64938ab3094cea1

mt-pref-appearance

bundles/net.tourbook/src/net/tourbook/messages.properties
```diff
@@ -3461,12 +3481,12 @@ merge_tour_source_graph_heartbeat_tooltip   = Merge heartbeat values from source
...
 
-pref_appearance_number_of_recent_tags         = Number of recent &tags:
+pref_appearance_number_of_recent_tags         = Number of recent &tags
 pref_appearance_number_of_recent_tags_tooltip = Number of recently used tags which are displayed in the context menu.\n\
                                                 \n\
                                                 0 will hide recently used tags.
-pref_appearance_pace_computation_option         = Pace computation options:
-pref_appearance_pace_computation_option_tooltip = This menu gives the ability to choose the time value to be used to compute and display pace values.
+pref_appearance_paceandspeed_computation_option         = Pace and Speed computation options
+pref_appearance_paceandspeed_computation_option_tooltip = This menu gives the ability to choose the time value to be used to compute and display pace and speed values.
```
wolfgang-ch, line 3489: The tooltip text `This menu gives...` references a menu but there is no a menu

wolfgang-ch: `And` in a action/group label should not be written in Uppercase this was the reason why I renamed `Pace And Speed Display`

When I'm not sure to write in upper- or lower-case. I'm looking at the Eclipse UI.

mt-label-uppercase

FJBDev: Good point, I will use this tip in the future

FJBDev:
> After the paused times are now working with my .fit data, it's a very cool feature :-)
> 
> Short pauses are now also visible in the timeline, without the displayed paused value, a pause could be hidden.
> 
> A problem is still to overlapped paused time which I already mentioned. When this should be solved, the tour marker labels should also be considered in the stacked labels which can be more complex because of the different tour marker label positions (vertical, horizontal)
> 
> mt-pauses

Yes, now that the FIT issues are fixed, I will work on showing stacked labels using the complex code of `net.tourbook.ui.tourChart.ValueOverlapChecker`

bundles/net.tourbook/src/net/tourbook/messages.properties
FJBDev: And I have just removed the "menu" text.

PR #239, opened 2020-10-24
Renaming from bikerWeight to bodyWeight when creating the Tour Data

FJBDev:
table

### DIFF
--- a/bundles/net.tourbook/src/net/tourbook/database/TourDatabase.java
+++ b/bundles/net.tourbook/src/net/tourbook/database/TourDatabase.java
@@ -2893,7 +2893,7 @@ public class TourDatabase {
             + " tourStartPlace      VARCHAR(" + TourData.DB_LENGTH_TOUR_START_PLACE + "),    \n" //$NON-NLS-1$ //$NON-NLS-2$
             + " tourEndPlace        VARCHAR(" + TourData.DB_LENGTH_TOUR_END_PLACE + "),      \n" //$NON-NLS-1$ //$NON-NLS-2$
             + " calories            INTEGER,                                                 \n" //$NON-NLS-1$
-            + " bikerWeight         FLOAT,                                                   \n" //$NON-NLS-1$
+            + " bodyWeight          FLOAT,                                                   \n" //$NON-NLS-1$
             + " " + KEY_BIKE + "    BIGINT,                                                  \n" //$NON-NLS-1$ //$NON-NLS-2$
 
             // from wolfgang


wolfgang-ch:
This exception occurs when no data option in the re-import dialog is selected -> The Re-import button should only be enabled when at least one option is selected

`java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at java.util.ArrayList.rangeCheck(ArrayList.java:659)
	at java.util.ArrayList.get(ArrayList.java:435)
	at net.tourbook.importdata.RawDataManager.actionReimportTour_10_Confirm(RawDataManager.java:740)
	at net.tourbook.importdata.RawDataManager.actionReimportTour(RawDataManager.java:588)
	at net.tourbook.ui.views.rawData.DialogReimportTours.doReimport(DialogReimportTours.java:474)
	at net.tourbook.ui.views.rawData.DialogReimportTours.access$17(DialogReimportTours.java:426)
	at net.tourbook.ui.views.rawData.DialogReimportTours$5.run(DialogReimportTours.java:575)
	at org.eclipse.swt.custom.BusyIndicator.showWhile(BusyIndicator.java:74)
	at net.tourbook.ui.views.rawData.DialogReimportTours.okPressed(DialogReimportTours.java:526)`

wolfgang-ch:
* When tours are selected, they should not be preloaded -> each tour should be loaded, reimported and saved and then the next tour should be loaded, ... to reduce the memory footprint

* When tours are preloaded and the Cancel button is pressed, then the reimport of the already preloaded tours is also performed -> When a Cancel button is pressed in any cases, then the whole process should be stopped

mt-preload-tours-2

wolfgang-ch:
When the option "All time slice data..." is selected then all other data options should be disabled except "Entire tour"

mt-reimport-dialog

FJBDev:
> This exception occurs when no data option in the re-import dialog is selected -> The Re-import button should only be enabled when at least one option is selected
> 
> `java.lang.IndexOutOfBoundsException: Index: 0, Size: 0 at java.util.ArrayList.rangeCheck(ArrayList.java:659) at java.util.ArrayList.get(ArrayList.java:435) at net.tourbook.importdata.RawDataManager.actionReimportTour_10_Confirm(RawDataManager.java:740) at net.tourbook.importdata.RawDataManager.actionReimportTour(RawDataManager.java:588) at net.tourbook.ui.views.rawData.DialogReimportTours.doReimport(DialogReimportTours.java:474) at net.tourbook.ui.views.rawData.DialogReimportTours.access$17(DialogReimportTours.java:426) at net.tourbook.ui.views.rawData.DialogReimportTours$5.run(DialogReimportTours.java:575) at org.eclipse.swt.custom.BusyIndicator.showWhile(BusyIndicator.java:74) at net.tourbook.ui.views.rawData.DialogReimportTours.okPressed(DialogReimportTours.java:526)`

Good catch, fixed

FJBDev:
> When the option "All time slice data..." is selected then all other data options should be disabled except "Entire tour"
> 
> mt-reimport-dialog

What's the difference between "All time slice data..." and "Entire tour" ?

FJBDev:
>     * When tours are selected, they should not be preloaded -> each tour should be loaded, reimported and saved and then the next tour should be loaded, ... to reduce the memory footprint

But isn't that removing a useful feature for the users ? Right now, selecting several tours (aka multiple tours) show all of them i the graphs... Won't the user miss this feature ?

wolfgang-ch:
In my screenshot above, first it would preload 3834 tours before it start the reimport for each single tour. I've not tested not this case because I canceled it but I think it may fail because of memory issues.

Instead of preloading at first all tours, it should load tour 1, reimport tour1, save tour 1, load tour2, reimport tour 2, save tour 2 and so on

FJBDev:
Ok, I think I understand this time.

I have modified my code so not all the tours are pre-loaded but instead, each tour is loaded, reimported and saved one at a time. It should be more efficient now.

Also, now, when the Cancel button is pressed, then the whole process is stopped

PR #242, opened 2020-10-30
 20.11 Test and fixes Part 2

FJBDev:
- We use the maxspeed to compute the max pace to clarify the code

- French translation


wolfgang-ch:
@FJBDev Should the reimport of the **paused time** not also set the **recorded time** ?

https://github.com/wolfgang-ch/mytourbook/blob/b69fbc13d021f77eeb6a1ab209246615aad95203/bundles/net.tourbook/src/net/tourbook/importdata/RawDataManager.java#L1078-L1080

FJBDev:
> @FJBDev Should the reimport of the **paused time** not also set the **recorded time** ?
> 
> https://github.com/wolfgang-ch/mytourbook/blob/b69fbc13d021f77eeb6a1ab209246615aad95203/bundles/net.tourbook/src/net/tourbook/importdata/RawDataManager.java#L1078-L1080

It is not necessary here as this is only used to display, in the log view, the difference between the previous paused time and the new paused time after re-import (since cloning a tour doesn't clone the pauses)

image

The recorded time and paused time is set [here](https://github.com/wolfgang-ch/mytourbook/blob/b69fbc13d021f77eeb6a1ab209246615aad95203/bundles/net.tourbook/src/net/tourbook/importdata/RawDataManager.java#L1463)

PR #244, opened 2020-11-05
20.11 Adding the ability to format the distance and distance difference values in the TourMarker view

FJBDev:
+ Misc code improvements


bundles/net.tourbook/src/net/tourbook/ui/views/geoCompare/GeoCompareView.java
```diff
@@ -966,8 +966,7 @@ private void compare_60_AllIsCompared(final GeoPartItem geoPartItem) {
          // this code may be compact and fast but I don't understand it
 
          final ArrayList<GeoPartComparerItem> filteredComparedTours = new ArrayList<>(geoPartItem.comparedTours.stream()
-               .filter(c -> _minDiffValueToRelative.apply(c.minDiffValue) >= 0 &&
-                     _minDiffValueToRelative.apply(c.minDiffValue) < _geoRelativeDifferencesFilter)
+               .filter(c -> isMinDiffValueWithinFilter(c.minDiffValue))
                .collect(Collectors.toList()));
```
wolfgang-ch, line 970: Now the 10 lines filter code is spread into 3 different code parts distributed all over the file.

This was the initial code which I can easily understand but now with 3 different code snippets it is hard for me.

mt-geo-compare

I checked the performance and it depends on the sequence which filter is run first. The first run of the stream() code is 5-10 times slower but further executions are similar like the old code, sometimes faster and sometimes slower.

mt-geo-compare-2

wolfgang-ch: I move the end time out of System.out.... and now the stream() code is about 5 times slower than the old code

mt-geo-compare-code

FJBDev: Interesting.... I thought they had designed the Stream in Java 8 to be more readable and faster.
I personally find the stream code more readable but I can revert it if you want. Let me know if you want me to put back the old code

wolfgang-ch: Because the stream() implementation is still below 1ms, we can ignore the "slow" performance but when the delay would be >10ms then we should optimize it -> we can keep it as a "bad" example :-)

I think, in this case, the stream() has some overhead why it is slower compared to a simple foreach loop.

FJBDev: Maybe what slows it down is the "new ArrayList<>"

I tend to think that new features are always better and faster but it's possible it's not the case.

PR #251, opened 2020-11-16
20.11 Device Paused Times fixes

FJBDev:
-  TCX Bugfix: When importing a file, we remove the invalid pauses (invalid time of 2007-04-01)

-  Fix: When the tour start time is modified, we need to update (offset) the paused start and end times

- :penguin: LINUX ONLY:penguin: : Fix for the database update message that is hidden below the splash message.

image

- Fix for body weight and fat statistics

- Misc french translation fixes


bundles/net.tourbook/src/net/tourbook/database/TourDatabase.java
```diff
@@ -4658,7 +4658,7 @@ private boolean updateDbDesign(final Connection conn, int currentDbVersion, fina
       final String[] buttons = new String[] { IDialogConstants.YES_LABEL, IDialogConstants.NO_LABEL };
 
       if ((new MessageDialog(
-            Display.getDefault().getActiveShell(),
+            splashManager.getShell(),
```
FJBDev, line 4661: @wolfgang-ch The solution i found is to grab the shell of the splash manager. What do you think?

wolfgang-ch: Interresting solution :-)

For the slideout + messagebox issue I tried several shells, but none worked only the on_top style

FJBDev: What is also interesting is that it was only an issue in Linux, not in Windows !?

PR #257, opened 2020-12-01
[BULK CHANGES]: Replacing all the .size() == 0 and > 0 calls with .isEmpty() for better performance

FJBDev:
Per this [rule](https://rules.sonarsource.com/java/tag/clumsy/RSPEC-1155), using `isEmpty()` is more readable and performant than using `.size()`

image


wolfgang-ch:
For me, these

```
.size() == 0 
.size() < 1
.size() <= 0 
```

replacements with `.isEmpty()` are OK


This replacement

```
.size() > 0
```

with `.isEmpty() == false` is for me less legible and need one more step to understand, also performance is propable the same.

Can you replace the `.isEmpty() == false` with the original code `.size() > 0`.

For me it would be OK when `.isEmpty() == false` could be replaced with `.hasContent()` but this is not available for collections.

FJBDev:
> with `.isEmpty() == false` is for me less legible and need one more step to understand, also performance is propable the same.
> 

I don't think the performance is the same. Example, if the collection has 2500 elements, the performance of .size() will be **O(2500)**
The performance of isEmpty()==false is **O(1)**

But I agree that it's not necessarily more readable
What about !arrayVariable.isEmpty() ?


> Can you replace the `.isEmpty() == false` with the original code `.size() > 0`.
> 
> For me it would be OK when `.isEmpty() == false` could be replaced with `.hasContent()` but this is not available for collections.

Another possibility would be to use the apache function [CollectionUtils.isNotEmpty()](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/CollectionUtils.html#isNotEmpty-java.util.Collection-)

Thoughts ?

wolfgang-ch:
> But I agree that it's not necessarily more readable
> What about `!arrayVariable.isEmpty() ?`
> 
> > Can you replace the `.isEmpty() == false` with the original code `.size() > 0`.
> > For me it would be OK when `.isEmpty() == false` could be replaced with `.hasContent()` but this is not available for collections.
> 
`.isEmpty() == false` and `!arrayVariable.isEmpty()` is for me a double negation "not no content" and I "hate" such constructions and this is the reason why I create https://github.com/wolfgang-ch/mytourbook/blob/2364f9274f1d0efbf6c3a21c3b6d433d0586fad1/bundles/net.tourbook.common/src/net/tourbook/common/util/StringUtils.java#L29-L31 some days ago


> Another possibility would be to use the apache function [CollectionUtils.isNotEmpty()](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/CollectionUtils.html#isNotEmpty-java.util.Collection-)
> And it also checks if it's not null so it can replace this
> 
> `if(arrayVariable != null && arrayVariable.size() > 0)`
> 
> with this
> 
> `if (CollectionUtils.isNotEmpty(arrayVariable))`
> 
> And actually, there is also a [CollectionUtils.isEmpty()](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/CollectionUtils.html#isEmpty-java.util.Collection-)
> 
> Thoughts ?


You can used these tools in your created code but I try to keep it simple especially when there are several && or || in an if statement. When it is very complex. sometimes I also split && and || into lines and comment each line or externalize it into a local vars. Since my bicycle accident I need to go the simple way because I cannot remember anymore too much in my short-term memory buffer.

FJBDev:
>>.isEmpty() == false and !arrayVariable.isEmpty() is for me a double negation "not no content" and I "hate" such constructions and this is the reason why I create 

That's true, a double negation is not very readable.

>>You can used these tools in your created code
Thanks but I try to avoid to have different standard depending on the creator of the file. I believe in consistency for a given project and I think that if you use a standard and I use a different standard, then it can be hard to go from one file to another if the code standard is different.

Therefore, I have done what you asked everywhere : "Can you replace the `.isEmpty() == false` with the original `code .size() > 0`."

I have created a [PR](https://github.com/wolfgang-ch/mytourbook/pull/258)

PR #258, opened 2020-12-03
Replacing the .isEmpty() == false with the original code .size() > 0

FJBDev:


wolfgang-ch:
I reverted one `.isEmpty() == false"` in https://github.com/wolfgang-ch/mytourbook/commit/7c6055e871396509c8daf1391f8826eed8295bef which I found im my review

mt-review-1

PR #260, opened 2020-12-07
Worldwind 2.2.0 Update

FJBDev:
The latest release of WorldWind 2.2.0 is requiring Java 11. Now that MT requires Java 11, we can update WorldWind

WorldWind 2.2.0 [Release notes](https://github.com/NASAWorldWind/WorldWindJava/releases/tag/v2.2.0) as of 08/12/2020


wolfgang-ch:
"New" 3D map is working :-)

...but it has the same issues as before, tour markers are still displayed when they shouldn't because they are behind the hill

mt-3d-map

mt-3d-map-2

FJBDev:
Hopefully it will be fixed in one of the next versions. Now that WorldWind is active again, we can hope for nice future features and bug fixes :-)

FJBDev:
[This issue](https://sourceforge.net/p/mytourbook/bugs/106/) is not fixed either but I already knew about :-(

PR #270, opened 2020-12-26
custom data tracks

gmngueko:
see: https://sourceforge.net/p/mytourbook/feature-requests/189/
This is the first part commit just for the data part, the UI part (timeslice view (already ready locally :-) )
but bigger will be done once this one is agreed by the community)

Change-Id: Idf5296a8bb03584e070a6843da26b0c9c44abebb
Signed-off-by: gmngueko <gmngueko@yahoo.com>

### DIFF
--- a/bundles/net.tourbook/src/net/tourbook/data/CustomTrackDefinition.java
+++ b/bundles/net.tourbook/src/net/tourbook/data/CustomTrackDefinition.java
@@ -1,0 +1,38 @@
+package net.tourbook.data;
+
+import java.io.Serializable;
+
+public class CustomTrackDefinition implements Serializable {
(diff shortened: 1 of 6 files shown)


wolfgang-ch:
Here are a few common findings

* Private fields should start with an _ underscore any other fields should NOT start with an _
* You should run the externalize string tool that texts are externalized or marked with `//$NON-NLS-1$`

Can you provide an example which data you want to put into custom data.
Currently this PR do not provide a possibility to enter or import data to test it

gmngueko:
ok I'll make another pull request with the extension I made on the FitlogEx importer from ST3. the extension allow the import

PR #279, opened 2021-01-03
Code modifications per previous PR's code suggestions

FJBDev:
See previous PR's code suggestions [here](https://github.com/wolfgang-ch/mytourbook/pull/277)


bundles/net.tourbook.cloud/src/net/tourbook/cloud/dropbox/PrefPageDropbox.java
```diff
@@ -337,4 +291,5 @@ private void updateTokensInformationGroup() {
       _labelExpiresAt.setEnabled(isAuthorized);
       _labelAccessToken.setEnabled(isAuthorized);
    }
+
```
wolfgang-ch, line 294: Authorization with the external browser is working in both cloud accounts :-)

...but there could be some improvements :-)

* The group names are different "Tokens Information" "Account Information"
* Sequence of the fields are different, "Refresh Token" and "Expires At" have not the same order, I would put "Expires At" at the bottom
* Like the strava pref page, the dropbox pref page could also have a webpage link

mt-cloud-dropbox


mt-cloud-strava

wolfgang-ch (commented):
almost done :-)

bundles/net.tourbook.common/src/net/tourbook/common/time/TimeTools.java
```diff
@@ -231,6 +231,14 @@
       };
    }
 
+   public static String constructLocalExpireAtDateTime(final long expireAt) {
+      if (expireAt == 0) {
+         return UI.EMPTY_STRING;
+      }
+
+      return Instant.ofEpochMilli(expireAt).atZone(TimeTools.UTC).format(DateTimeFormatter.ISO_DATE_TIME);
```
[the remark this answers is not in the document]
FJBDev: Fixed in next PR

bundles/net.tourbook.cloud/src/net/tourbook/cloud/dropbox/PrefPageDropbox.java
FJBDev: >The group names are different "Tokens Information" "Account Information"

They are called different because the Strava one contains the athlete's name and athlete's web page whereas the Dropbox ONLY contains token information.
Let me know what you prefer me to rename it to.

> Sequence of the fields are different, "Refresh Token" and "Expires At" have not the same order, I would put "Expires At" at the bottom
Fixed in next PR

> Like the strava pref page, the dropbox pref page could also have a webpage link
Dropbox doesn't offer a "public" user webpage. So are you saying you want to see just a plain https://www.dropbox.com link ?

wolfgang-ch: > > The group names are different "Tokens Information" "Account Information"
> 
> They are called different because the Strava one contains the athlete's name and athlete's web page whereas the Dropbox ONLY contains token information.
> Let me know what you prefer me to rename it to.

I would use a more common group name, e.g.

* Cloud Access
* Cloud Account
* Cloud Information

wolfgang-ch: > > Like the strava pref page, the dropbox pref page could also have a webpage link
> 
> Dropbox doesn't offer a "public" user webpage. So are you saying you want to see just a plain https://www.dropbox.com link ?

Yes, just this link with a label like e.g. "Web page"

Only the first word in a label (Not Actions) should start with an uppercase any other in lowercase, I know this is not everywere in MT but since a while I use this rule

FJBDev: >I would use a more common group name, e.g.

Fixed in my next PR

>Yes, just this link with a label like e.g. "Web page"

Fixed in my next PR

>Only the first word in a label (Not Actions) should start with an uppercase any other in lowercase

Fixed in my next PR for both the Dropbox and Strava preference pages

PR #280, opened 2021-01-03
custom tracks #6

gmngueko:
correct styling issues

PR #281, opened 2021-01-05
custom tracks #6c

gmngueko:
correcting some more styling issues


gmngueko:
PR #280  was updated correctly, so this one is not needed (and actually should have been done on "contributor-gmngueko" branch anyway :-)

PR #282, opened 2021-01-06
Cloud Preference pages improvements

FJBDev:
Per code review [suggestions](https://github.com/wolfgang-ch/mytourbook/pull/279#discussion_r551993714)

- Renamed the group to "Cloud Account"
- Added the Dropbox web page link
- Changed all the labels so that the first word in a label should start with an uppercase any other in lowercase

image

image

- Removed slf4j warnings

Before fix
image

After fix
image


bundles/net.tourbook.cloud/src/net/tourbook/cloud/dropbox/messages.properties
```diff
@@ -11,9 +11,10 @@ Image__Dropbox_Folder       = dropbox-folder.png
...
-Pref_CloudConnectivity_Dropbox_ExpiresAt_Label          = Expires At
-Pref_CloudConnectivity_Dropbox_RefreshToken_Label       = Refresh Token
-Pref_CloudConnectivity_Dropbox_Tokens_Information_Group = Tokens Information
+Pref_CloudConnectivity_Dropbox_AccessToken_Label   = Access token
+Pref_CloudConnectivity_Dropbox_AccessToken_Tooltip = This is the access token to access your Dropbox account
+Pref_CloudConnectivity_Dropbox_Button_Authorize    = Authorize
+Pref_CloudConnectivity_Dropbox_ExpiresAt_Label     = Expires at
+Pref_CloudConnectivity_Dropbox_RefreshToken_Label  = Refresh token
+Pref_CloudConnectivity_Dropbox_WebPage_Label       = Web page
+Pref_CloudConnectivity_Dropbox_WebPage_Link        = https://www.dropbox.com
```
wolfgang-ch, line 20: FYI and not an issue.

I'm using a little bit different naming structure for the text keys, a the name of the used control (or other type) is the 3rd part in the name and not at the end. At the end there is e.g. _Tooltip, _Message or _Title, here is an example:

mt-text-key-naming

bundles/net.tourbook.cloud/src/net/tourbook/cloud/dropbox/PrefPageDropbox.java
```diff
@@ -144,9 +148,25 @@ private void createUI_20_TokensInformation(final Composite parent) {
 
       _group = new Group(parent, SWT.NONE);
       GridDataFactory.fillDefaults().grab(true, false).applyTo(_group);
-      _group.setText(Messages.Pref_CloudConnectivity_Dropbox_Tokens_Information_Group);
+      _group.setText(net.tourbook.cloud.Messages.Pref_CloudConnectivity_CloudAccount_Group);
       GridLayoutFactory.swtDefaults().numColumns(2).applyTo(_group);
       {
```
wolfgang-ch, line 153: I think that once I've already described this problem but I don't know where it was.

When using externalized strings from more than one `Messages.java` files, then the externalize string tool **do not work.**

mt-externalize-strings

My solution is, which you can find in many places, extract strings from the 2nd and further Messages.java into constants and put them at the top of the file (so you can find them easier)

mt-externalize-strings-from-other-locations


When you want to run the externalize string tool, first comment out these constants at the top (which causes an error), run the externalize string tool, after that revert the comments, someday I found this handy solution.

bundles/net.tourbook.common/src/net/tourbook/common/time/TimeTools.java
```diff
@@ -232,10 +231,6 @@
    }
 
    public static String constructLocalExpireAtDateTime(final long expireAt) {
```
wolfgang-ch, line 233: `createISODateTime` as method name ?

TimeTools.java contains common methods, so `constructLocalExpireAtDateTime` is a little bit specific

FJBDev: Good idea, I'll do that in my next PR

bundles/net.tourbook.cloud/src/net/tourbook/cloud/dropbox/PrefPageDropbox.java
FJBDev: Oh I forgot about that. I will address that in my next PR

bundles/net.tourbook.cloud/src/net/tourbook/cloud/dropbox/messages.properties
FJBDev: I wouldn't mind if you wanted to establish a standard rule such as :
Pref_ClassName_CompositeName_UIType ?

bundles/net.tourbook.cloud/src/net/tourbook/cloud/dropbox/PrefPageDropbox.java
FJBDev: >When you want to run the externalize string tool, first comment out these constants at the top (which causes an error), run the externalize string tool, after that revert the comments, someday I found this handy solution.

It's a very "hacky" solution though, I am not a fan of those "hacks"

bundles/net.tourbook.common/src/net/tourbook/common/time/TimeTools.java
FJBDev: Renamed to `getUTCISODateTime()`

bundles/net.tourbook.cloud/src/net/tourbook/cloud/dropbox/messages.properties
wolfgang-ch: I think it is difficult to create a common rule but the text keys are automatically grouped by the first two parts which are separated with an _ (underscrore), this is the reason why images have 2 subsequent underscrores.

In the beginning I started words in the text keys with lowercase but they cannot be navigated easily with the cursor keys (depending on the location in the IDE) so I switched to ProperCase.

bundles/net.tourbook.cloud/src/net/tourbook/cloud/dropbox/PrefPageDropbox.java
wolfgang-ch: The other solution would be to improve the string externalisation tool which is propably very difficult, you have to manage different Messages.java files.

PR #285, opened 2021-01-08
 Adding the ability to upload manual tours to Strava

FJBDev:
- Adding the ability to upload tours that don't contains time series
Because Strava gives the activity Id directly back, the user can click on the Strava link in the LogView to open the uploaded activity in their system's browser:

image


bundles/net.tourbook.common/src/net/tourbook/common/time/TimeTools.java
```diff
@@ -569,6 +565,10 @@ public static int getTimeZoneIndex_WithDefault(final String timeZoneId) {
       return tzIndex;
    }
 
+   public static String getUTCISODateTime(final long date) {
+      return Instant.ofEpochMilli(date).atZone(TimeTools.UTC).format(DateTimeFormatter.ISO_DATE_TIME);
+   }
+
```
wolfgang-ch, line 571: :-)

bundles/net.tourbook.cloud/src/net/tourbook/cloud/strava/PrefPageStrava.java
```diff
@@ -208,7 +208,7 @@ public void widgetSelected(final SelectionEvent e) {
    }
```
wolfgang-ch, line 209: The port validation works :-)

but when this port is occupied by any other app then we have a problem, however when the port must be static by the cloud provider then this problem can not solved easily, time will tell.

mt-pref-page-strava

wolfgang-ch: I discovered this issue:

* Assume any accounts are reset (may be not necessary)
* Select Dropbox pref page
* Presse Authorize
* Select Strava pref page
* Press Connect...
-> The port is blocked, strava cannot be connected

My solution could be:
When a user is leaving Dropbox/Strava pref page (I think `performCancel()` is called) then the user should be informed that the connection is closed and the connection must be closed

However when this feature is used "correctly" then it is very simple to be used with the cached user id/password :-)

wolfgang-ch (commented):
one remaining issue

bundles/net.tourbook.cloud/src/net/tourbook/cloud/strava/PrefPageStrava.java
FJBDev: I thought about this case but it seems to me that it's a very specific "corner" case, meaning it should probably not happen.
So is it worth adding code for this case that has a 1% chance happening ?

FJBDev: > 
> 
> The port validation works :-)
> 
> but when this port is occupied by any other app then we have a problem, however when the port must be static by the cloud provider then this problem can not solved easily, time will tell.
> 
Yes, time will tell. Hopefully the port number is "random" enough so that it will not cause conflicts with user's computers. But at this time, Strava and Dropbox don't offer a dynamic port callback solution

wolfgang-ch: > I thought about this case but it seems to me that it's a very specific "corner" case, meaning it should probably not happen.
> So is it worth adding code for this case that has a 1% chance happening ?

I did more testing and it happened that after closing the pref dialog and reopening it, the port error message appeared immediately after pressing the connect button the first time -> port was not closed

So we should add this code on each cloud pref page

```
@Override
   public boolean okToLeave() {

      if (_server != null) {
         _server.stopCallBackServer();
      }

      return super.okToLeave();
   }
```

FJBDev: Good idea, I will add that to the code.
But I think if the users follow well the documentation I made in my [PR](https://github.com/wolfgang-ch/mytourbook/pull/273), it should go smoothly. Plus, it should be a one-time operation.

wolfgang-ch: >But I think if the users follow well the documentation 

Who is reading a documentation, 1%?

FJBDev: Hopefully the users are reading the documentation 😄

PR #287, opened 2021-01-11
TCX Export Enhancement and fixes

FJBDev:
Per this [feature request](https://sourceforge.net/p/mytourbook/feature-requests/144/)

- Added the speed and power values in the TCX exports (updated unit tests)

image

- Fixed the TCX Course Export format as it was not properly formatted before and was preventing from importing TCX Courses in Garmin Connect (updated unit tests)

When using a TCX Course exported with MT 20.12.0:

When the TCX is generated with the code fix, the course is now imported successfully in [Garmin Connect Courses](https://connect.garmin.com/modern/courses):

- Enhanced the unit tests so that they handle the case where the user does or does not have vendor credentials.

Unit tests 100% successful
image


bundles/net.tourbook.cloud/src/net/tourbook/cloud/dropbox/PrefPageDropbox.java
```diff
@@ -226,6 +226,16 @@ public void init(final IWorkbench workbench) {
...
 
+   @Override
+   public boolean okToLeave() {
+
+      if (_server != null) {
+         _server.stopCallBackServer();
+      }
+
+      return super.okToLeave();
+   }
```
wolfgang-ch, line 237: I had still a case when the port was blocked but it occured after I clicked many times on the authorized button and using an old instance of the opened web page, I think this is OK. 

When this tool is used as a normal user then it works as expected.

wolfgang-ch:
The Unit test has 4 errors

[net.tourbook.tests 20210113-152814.xml.zip](https://github.com/wolfgang-ch/mytourbook/files/5809174/net.tourbook.tests.20210113-152814.xml.zip)

FJBDev:
> The Unit test has 4 errors
> 
> [net.tourbook.tests 20210113-152814.xml.zip](https://github.com/wolfgang-ch/mytourbook/files/5809174/net.tourbook.tests.20210113-152814.xml.zip)

I just realized Eclipse still had 1 modified file not commited. Fixed in new PR

bundles/net.tourbook.cloud/src/net/tourbook/cloud/dropbox/PrefPageDropbox.java
FJBDev: I knew that case too but a user shouldn't click several times on authorize. But sometimes users make mistake....

PR #293, opened 2021-01-23

FJBDev:
See this [PR](https://github.com/wolfgang-ch/mytourbook/pull/292)'s code review


wolfgang-ch:
In this case, the 3rd text key part should be the UI control name, the text `&User-Agent` is for a label and not a text control

mt-label

wolfgang-ch:
This modification can cause a NPE, default for weather cloud is null

mt-possible-NPE

bundles/net.tourbook/src/net/tourbook/map2/action/ActionExportMapViewClipboard.java
```diff
@@ -42,14 +43,20 @@ public ActionExportMapViewClipboard(final Map2View mapView) {
...
+      final ImageData imageData = mapViewImage.getImageData();
+
+      if (UI.IS_LINUX) {
+         final PngTransfer imageTransfer = PngTransfer.getInstance();
+         clipboard.setContents(new Object[] { imageData },
+               new Transfer[] { imageTransfer });
+      } else {
+         final ImageTransfer imageTransfer = ImageTransfer.getInstance();
+         clipboard.setContents(new Object[] { imageData },
+               new Transfer[] { imageTransfer });
```
wolfgang-ch, line 58: The map is a little bit empty when pasting into gimp or inkscape

mt-copy-map-to-clipboard

FJBDev: I used MS Paint and it worked, can you try and tell me? 
There might be a transparent layer issue with the set code

FJBDev:
> In this case, the 3rd text key part should be the UI control name, the text `&User-Agent` is for a label and not a text control
> 
> mt-label

I will fix.
In this case its a necessary change because in java the == sign should not be used to compare stringd

FJBDev:
> In this case, the 3rd text key part should be the UI control name, the text `&User-Agent` is for a label and not a text control
> 
> mt-label

Fixed
image

FJBDev:
> This modification can cause a NPE, default for weather cloud is null
> 
> mt-possible-NPE
> 
> mt-possible-NPE-2

Fixed by using [Object.equals()](https://stackoverflow.com/a/50891742/7066681)

bundles/net.tourbook/src/net/tourbook/map2/action/ActionExportMapViewClipboard.java
FJBDev: This is what I see in Windows:
- When pasted in MS Paint, it works fine
- When pasted in GIMP, it shows a transparent map
image

wolfgang-ch: * MS Paint is OK
* Inkscape 
 ** is OK when using `net.tourbook.map2.view.PngTransfer`
** is not OK when using `org.eclipse.swt.dnd.ImageTransfer`

FYI: `org.eclipse.swt.graphics.ImageData.transparentPixel` is used in the 2D map

mt-map-image-clipboard

FJBDev: > 
> 
>     * MS Paint is OK
> 
>     * Inkscape
>       ** is OK when using `net.tourbook.map2.view.PngTransfer`
>       ** is not OK when using `org.eclipse.swt.dnd.ImageTransfer`

Should we then ask the user to only use MS Paint ? That seems a bit restrictive.
I found [this bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=299138) that seems similar but the solution uses awt.

> 
> FYI: `org.eclipse.swt.graphics.ImageData.transparentPixel` is used in the 2D map
> 
I tried but it seems that the transparentPixel is not used
image

FJBDev: After more investigation, I have found a fix for this issue 😄 
Now, exporting the map as an image to the clipboard it works both on Linux and Windows.
I have created a PR

image

wolfgang-ch: 🥇 

I played a little bit with the transparent pixel but was not successful

FJBDev: For now it works good but maybe with time, a better solution will be found.

PR #286, opened 2021-01-11
custom tracks series data views

gmngueko:
implementation of time slice view
 and graph chart view for custom tracks
 series data

PR #296, opened 2021-01-26
custom tracks final: correct cosmetics

gmngueko:
correct cosmetics see: https://github.com/wolfgang-ch/mytourbook/pull/290

cust-track-2021-01-24_A
cust-track-2021-01-24_B
cust-track-2021-01-24_C
cust-track-2021-01-24_D
 

add rest and lap note in FitlogEx import


FJBDev:
@gmngueko 
@wolfgang-ch  has merged my recent work to main and it includes the fix for the FitLog & FitLogEx unit tests. 
You might want to look at the [commit](https://github.com/wolfgang-ch/mytourbook/commit/819e61c2dd56c9a4341d764fbf429369dd2715bc) and merge with main to update the FitLog & FitLogEx unit tests with your work.

bundles/net.tourbook.common/src/net/tourbook/common/util/ColumnManager.java
```diff
@@ -1441,8 +1441,8 @@ private void fillHeaderCtxMenu_40_Columns(final Menu contextMenu) {
        */
       final ArrayList<ColumnDefinition> allColumns = getRearrangedColumns();
 
-      final ArrayList<ColumnDefinition> displayedColumns = new ArrayList<>();
-      final ArrayList<ColumnDefinition> notDisplyedColumns = new ArrayList<>();
+      final ArrayList<ColumnDefinition> displayedColumns = new ArrayList<ColumnDefinition>();
```
FJBDev, line 1444: Isn't that extraneous ? See rule [2293](https://rules.sonarsource.com/java/RSPEC-2293)

gmngueko: > Isn't that extraneous ? See rule [2293](https://rules.sonarsource.com/java/RSPEC-2293)

I didn't touch this code actually.
on top of that you and wolfgang should talk because he is advocating to have the type in <> and you are advocating for the reverse.

FJBDev: Oh yes, that's what I don't like when doing PRs with branches that are not up to date with main because then you don't know if the file change is in the PR or because the destination branch is not updated with the latest main, really annoying.

Well, if @wolfgang-ch prefers to have the type in both <> then I suppose we should do that, I was not aware of it.
Personnally, i don't see the purpose because it's redundant to have it defined twice and would prefer to follow the Java Standard practice rule. But that's my opinion

gmngueko: > Oh yes, that's what I don't like when doing PRs with branches that are not up to date with main because then you don't know if the file change is part of the PR or if it's appearing because the destination branch is not updated with the latest main, really annoying.
> I've had cases where I modified 2 files but yet more than 50 files appear in the PR because of that. When going back to the PR then we dont know what was intended to be modified and it creates a bunch of "noise" and possible confusion and at times headaches
> 
> Well, if @wolfgang-ch prefers to have the type in both <> then I suppose we should do that, I was not aware of it.
> Personally, i don't see the purpose because it's redundant to have it defined twice and would prefer to follow the Java Standard practice rule. But that's just my opinion.

Agree with you, especially as the save action preference mentionned in the MTB standard setting says specifically "Remove redundant type arguments" which exactly does what you are advocating. But apparently that settings doesn't behave that way on wolfgang setup. 
At this point I'am oblige to revert back, on my code, with "ctrl+z" to prevent that "save action" rule  to remove the type !!....But ok, at this point I just want the "custom tracks" to be finished/merged in Main, so I can start something else.

FJBDev: I personally don't mind if we need to use diamond operator or not as longs as it's consistent.

wolfgang-ch: > you and wolfgang should talk because he is advocating to have the type in <> and you are advocating for the reverse( no type in <>).

I cannot remember advocating to have the type in the <> in the constructor, some days ago I just complained that the yellow flag should not appear, which is the case when you have an empty <> in the constructor. 

As I remember the empty <> was missing in the constructor.

I found the discussion https://github.com/wolfgang-ch/mytourbook/pull/286#discussion_r559177528

bundles/net.tourbook/src/net/tourbook/ui/tourChart/SlideoutTourChartGraphs.java
```diff
@@ -18,6 +18,7 @@
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
+import java.util.Map;
```
wolfgang-ch, line 22: This review is for the whole slideout, so it is not specific to this line.

First of all you have implemented a separate UI when custom data are availale, just the way I wanted 👍 

...but there are some issues

wolfgang-ch, line 22: I discovered that showing the slideout with custom data is very slow, so I profiled it and it takes almost 1 second to layout 52 toolbars. 

I use this profiling tool for MT since 10+ years to find bottlenecks when they are discovered.

mt-profiling

wolfgang-ch: I'm not happy with the slideout UI, it's just lake a maze to find a graph, there is no sorting or categorzation. 

I would use a column format
* 1st column contains all already existing graphs
* 2nd .. nth column contains the custom track graph

For this screenshot, I removed the inner border and resized the slideout to show ALL graphs

mt-all-chart-graphs


The graph min/max slideout is already showing the existing graphs in a column and the displaying time is much faster compared with the graph slideout with custom tracks

mt-graph-min-max

wolfgang-ch: The columns are also more or less randomly sorted, not very user friendly

mt-cust-track-columns

wolfgang-ch (commented):
Before I review the code, first the UI should be OK but I already discovered some shortcut names, e.g. `cnt` which I don't like

bundles/net.tourbook/src/net/tourbook/ui/tourChart/SlideoutTourChartGraphs.java
gmngueko: @wolfgang-ch  thanks. Does your profiler gives a hint on why the code is slow, meaning what could be done to make it faster ?
1) I'll modify the display to be in Alphabetical Order
2) I'll investigate how to transform in "Column Format"
3) change 'cnt' to 'count'

bundles/net.tourbook.common/src/net/tourbook/common/util/ColumnManager.java
gmngueko: > > you and wolfgang should talk because he is advocating to have the type in <> and you are advocating for the reverse( no type in <>).
> 
> I cannot remember advocating to have the type in the <> in the constructor, some days ago I just complained that the yellow flag should not appear, which is the case when you have an empty <> in the constructor.
> 
> As I remember the empty <> was missing in the constructor.
> 
> I found the discussion [#286 (comment)](https://github.com/wolfgang-ch/mytourbook/pull/286#discussion_r559177528)

@wolfgang-ch , sorry, it seems I misunderstood your review I thought it was about specifying the type, now I got it. it was simply the missing <> in the constructor which was the issue.

bundles/net.tourbook/src/net/tourbook/ui/tourChart/SlideoutTourChartGraphs.java
wolfgang-ch: >Does your profiler gives a hint on why the code is slow, meaning what could be done to make it faster ?

No, however you can see the code location where it occurred

>     3. change 'cnt' to 'count'

`count` for what ?

Since a while I use the prefix `num` (reading: "number of") for fields which are counting something, e.g.

mt-num

gmngueko:
> @gmngueko
> @wolfgang-ch has merged my recent work to main and it includes the fix for the FitLog & FitLogEx unit tests.
> You might want to look at the [commit](https://github.com/wolfgang-ch/mytourbook/commit/819e61c2dd56c9a4341d764fbf429369dd2715bc) and merge with main to update the FitLog & FitLogEx unit tests with your work.

@FJBDev  unit test merge done with my work in https://github.com/wolfgang-ch/mytourbook/pull/297

PR #308, opened 2021-02-26
Misc fixes that slipped through the cracks

FJBDev:


bundles/net.tourbook.cloud/src/net/tourbook/cloud/messages.properties
```diff
@@ -1,103 +1,89 @@
...
+PrefPage_CloudConnectivity_Label_WebPage               = Web page
+PrefPage_CloudConnectivity_UnavailablePort_Message     = The port {0} necessary to retrieve the {1} token is unavailable or already in use.
+PrefPage_CloudConnectivity_UnavailablePort_Title       = Port Unavailable
+
+PrefPage_SuuntoWorkouts_Checkbox_SinceDateFilter = &Use a date filter
+PrefPage_SuuntoWorkouts_FolderPath_Tooltip       = This is the folder where the new workout files will be downloaded to
+PrefPage_SuuntoWorkouts_Label_FolderPath         = Download &folder
+PrefPage_SuuntoWorkouts_SinceDateFilter_Tooltip  = Only workouts for which the start date is greater than this date will be downloaded
+
+Suunto_WorkoutsDownloader_Tooltip = Download workouts from a Suunto App account
```
wolfgang-ch, line 84: `Suunto_WorkoutsDownloader_Tooltip`

All other text keys in the same dialog have this prefix `Dialog_DownloadWorkoutsFromSuunto_...`

bundles/net.tourbook.cloud/src/net/tourbook/cloud/messages.properties
```diff
@@ -1,103 +1,89 @@
...
+Dialog_UploadToursToStrava_SubTask = {0} Tour Preparation\n\
+                                     {1} Upload
+Dialog_UploadToursToStrava_Task    = Uploading {0} tours to Strava ({1})
+Dialog_UploadToursToStrava_Title   = Strava Tour Upload Summary
 
-Suunto_Workouts_Description = Download workouts from a Suunto App account
+Dialog_ValidatingSuuntoTokens_SubTask = \u231B Validating the Suunto App tokens
 
-ValidatingSuuntoTokens_SubTask = \u231B Validating the Suunto App tokens
+Html_CloseBrowser_Text = You can now close this browser's window
```
wolfgang-ch, line 32: `Html_CloseBrowser_Text`

A better name would be `Html_Text_CloseBrowser` for the possible case that additional `Html_Text_...`  text keys are needed

wolfgang-ch (commented):
Is now merged with 2 minor issues

bundles/net.tourbook.cloud/src/net/tourbook/cloud/messages.properties
FJBDev: This one is different because it is not part of a dialog. It is the tooltip for the Easy Import Manager entry, as below:

Let me know if you think of a better name and I will change it

image

bundles/net.tourbook.cloud/src/net/tourbook/cloud/messages.properties
FJBDev: Fixed

bundles/net.tourbook.cloud/src/net/tourbook/cloud/messages.properties
wolfgang-ch: >This one is different because it is not part of a dialog. 

I've not investigated any detail, I just saw it as a single text key. When this text is related to the easy import launcher page, then the text key should be similar like the other text keys

https://github.com/wolfgang-ch/mytourbook/blob/243549e5e3a21dbb8a0563928add6f479b2bb0ab/bundles/net.tourbook/src/net/tourbook/messages.properties#L1373-L1378

FJBDev: Ok, I will use the same name pattern then

FJBDev: Fixed in this [PR](https://github.com/wolfgang-ch/mytourbook/pull/310)

bundles/net.tourbook.cloud/src/net/tourbook/cloud/messages.properties
FJBDev: Fixed in this [PR](https://github.com/wolfgang-ch/mytourbook/pull/310)

PR #306, opened 2021-02-25
Tour Chart Inverted Pace

FJBDev:
Currently, MyTourbook shows the pace graph with descending values : The faster the user is running, the lower the value is and the slower the user is running, the higher the value is.
This is not intuitive because it is essentially an "inverted" graph when we compare it to the speed graph and makes it harder for the user to "read" the graph.

image

Example with Strava where the pace is correctly displayed:
image

After the [fix](https://github.com/wolfgang-ch/mytourbook/pull/306/commits/c64090fcdcf6abcf8d426cc0f7d1c1d517471169):
image

- Misc code improvements found when I tested my code : 
  - [Array designators "[]" should be located after the type in method signatures](https://rules.sonarsource.com/java/tag/convention/RSPEC-1195)
  - [Removing unused parameters](https://rules.sonarsource.com/java/tag/unused/RSPEC-1172)
  - [Making anonymous inner class a lambda](https://rules.sonarsource.com/java/RSPEC-1604)

PR #310, opened 2021-03-01
Tour Charted Inverted pace without code improvements

FJBDev:
This PR reverts ALL my code improvements in this previous [PR](https://github.com/wolfgang-ch/mytourbook/pull/306)

If you compare with main, you will only see the useful code to address the pace chart issue.


wolfgang-ch (commented):
Now reviewing this PR took me 5 minutes.

There are 2 things which should be improved:

* The horizontal sliders are swapped when dragged, which I already mentioned in a comment but there were not yet any complain, now it's easier to discover

mt pace horizontal slider

* The user should have an option to select the direction where high/low values are displayed because this is a big change and maybe not all users like it. You can place the UI below the heartbeat and group them into a new group "Graphs"

mt pace in chart options

wolfgang-ch:
I just discovered that the slider value and scale value are not the same


mt pace value vs scale 2

FJBDev:
>     * The horizontal sliders are swapped when dragged, which I already mentioned in a comment but there were not yet any complain, now it's easier to discover

I will investigate this
> 
>     * The user should have an option to select the direction where high/low values are displayed because this is a big change and maybe not all users like it. You can place the UI below the heartbeat and group them into a new group "Graphs"
> 
 Which one would you prefer by default ?

FJBDev:
> I just discovered that the slider value and scale value are not the same


Good catch, I will investigate this

wolfgang-ch:
> > * The user should have an option to select the direction where high/low values are displayed because this is a big change and maybe not all users like it. You can place the UI below the heartbeat and group them into a new group "Graphs"
> >
> 
> Which one would you prefer by default ?

What is more common?

We can use the more common as default and have to write into the readme `Changes` section when it's changed

FJBDev:
> What is more common?

I would tend to think that displaying the pace like the speed is more common but nobody has complained all this time.
Other software like Strava do that and now that I know of this bug I wouldn't use the previous display any more but I don't know for the long time users.

FJBDev:
> I just discovered that the slider value and scale value are not the same
> 
I have fixed this issue. 

image

FJBDev:
>     * The horizontal sliders are swapped when dragged, which I already mentioned in a comment but there were not yet any complain, now it's easier to discover
> 
I have fixed this issue as well:

image
image


When we have decided if we want to offer the option to the user to display the pace chart in inverted and non inverted, which one by default, or if we display in inverted only, I will do a new PR.

Also, I think we should invert those valued, what do you think ?

image

wolfgang-ch:
> Also, I think we should invert those values, what do you think ?

It should depend which top/bottom value the user has selected in the new tour chart options for the pace graph.

I just see in this dialog, that the checkboxes are not vertically aligned in the middle.

FJBDev:
> I just see in this dialog, that the checkboxes are not vertically aligned in the middle.

I can fix that

FJBDev:
Oh actually, I see you already fixed it [here](https://github.com/wolfgang-ch/mytourbook/commit/1d91b49c25c357c388142c132fa0ca7d36986a0c)

FJBDev:
Also, the spinners are not vertically aligned

image

wolfgang-ch:
> 
> 
> Also, the spinners are not vertically aligned
> 
> image

You mean horizontally aligned but this is not possible when each spinner has different min/max values. It looks more ugly when the +/- in the spinner control are not in one line (left aligned).

FJBDev:
What about like this ? Does it look better ?

image
I modified it in my new [PR ](https://github.com/wolfgang-ch/mytourbook/pull/311)but let me know if you don't like it and I'll revert it.

wolfgang-ch:
> Does it look better ?

Fore sure :-)

I thought the width could not be set but you've done it with the layout

I just tested it with different min/max values and the spinner width is adjusted accordingly. Power max is 1'000'000, when you reduce it to 100'000 the width of the slideout is smaller

wolfgang-ch:
> 
> 
> > I just discovered that the slider value and scale value are not the same
> 
> I have fixed this issue.
> 
> image

It works now as it should and also the _Daytime_ statistic :-)

It was not much code changes you did but for me it is relative complex code with all the different variables.

FJBDev:
> It was not much code changes you did but for me it is relative complex code with all the different variables.

Indeed, it was little code modification but it was hard to find and took time to test. Sometimes little code changes can have big impact

PR #313, opened 2021-03-05
21.3 Misc changes

FJBDev:
- 21.3 french translation fixes
- Removed warnings in several files such as unused imports, unused parameters


bundles/net.tourbook.common/src/net/tourbook/common/messages_it.properties
```diff
@@ -175,6 +175,7 @@ Image__MapProvider_Profile          = map-provider-profile.png
 Image__MapProvider_Profile_Hill     = map-provider-profile-hill.png
 Image__MapProvider_Transparent      = map-provider-transparent.png
 Image__MapProvider_Transparent_Hill = map-provider-transparent-hill.png
+Image__Options_Dark                 = options-dark.png
```
FJBDev, line 178: @wolfgang-ch 
In your review [comment](https://github.com/wolfgang-ch/mytourbook/pull/301#discussion_r581209308), you said this at the time:
_"I'm currently not happy that all images are defined in messages.properties and these entries are duplicated in 99.99% cases for all languages."_

Are we still supposed to duplicate the images for all languages ?

wolfgang-ch: >Are we still supposed to duplicate the images for all languages ?

I'm investigating this but that all translation errors do disappear from the error view, I set all images to be "translated".

Accidentally I also commited this test file https://github.com/wolfgang-ch/mytourbook/blob/f7efc93fd5fb85e3a1022dfe0e410b0effaaf530/bundles/net.tourbook/src/net/tourbook/map2/Images.java which is not yet used and not final, it was just a test.

When replacing translated images into constants, it will modify many files so I will not do it for 21.3

wolfgang-ch:
Can you revert the modifications which are not necessary. I don't want to spend more time into reviews which are not necessary.

There are many hundreds of yellow flags in MT, but MT is stable. 
So there is no need to change the code just that code policies are happy.

I also tried code optimizations 10+ years ago and code policies will never be happy, depending on the setup they will find always things which **"MUST"** be modified when you have a large code base.

FJBDev:
> Can you revert the modifications which are not necessary. I don't want to spend more time into reviews which are not necessary.
> 
> There are many hundreds of yellow flags in MT, but MT is stable.
> So there is no need to change the code just that code policies are happy.
> 
> I also tried code optimizations 10+ years ago and code policies will never be happy, depending on the setup they will find always things which **"MUST"** be modified when you have a large code base.

Sure. I just reverted the modifications which are not necessary

bundles/net.tourbook.common/src/net/tourbook/common/messages_it.properties
FJBDev: Ok. I think that is why I also translated the images for the french files when I started contributing to MT because it was showing as an error.

The Images.java file is a good idea but for sure, it will take time and require a lot of modifications

wolfgang-ch: > 
> 
> Ok. I think that is why I also translated the images for the french files when I started contributing to MT because it was showing as an error.
> 
> The Images.java file is a good idea but for sure, it will take time and require a lot of modifications

Normally the translators do not need to "translate" images because I copy the image name into all language files before they start the translation, so they never see the image errors. Because you are not a "normal" translator, you are seeing the image errors.

When I copy the image name in all languages, then I see that the messaging editor gets really slow. Some time I'll spend time to speed it up when possible but you need a lot of internal Eclipse IDE knowhow. This is also a reason to extract the image names from the translatable texts.

PR #314, opened 2021-03-08
Fix for the segmenter pace graph when the display is inverted & small code enhancements

FJBDev:
Inverted Pace Graph:
image

Non-Inverted Pace Graph:
image

Fixes after testing:
- When uploading a tour to suunto app that doesn't have time data, we need to set a very low speed so that all the track points are generated otherwise the gpx file is not complete
- We log the response payload in case the tokens are not successfully retrieved
- When the user clicks on Cancel (Tour Exporter, Map2D Image Exporter) when a file already exists, the dialog should not be closed


wolfgang-ch:
> Inverted Pace Graph:
> image
> 

I'm wondering why your negative elevation values are black and not green as they should

mt-graph-segments

FJBDev:
> I'm wondering why your negative elevation values are black and not green as they should

I didn't notice that. The screenshots I did were with a new .metadata and derby-database folders, not with my production data. If you try with a fresh configuration, do you see the black colors like my screenshots ?

wolfgang-ch:
> 
> 
> > I'm wondering why your negative elevation values are black and not green as they should
> 
> I didn't notice that. The screenshots I did were with a new .metadata and derby-database folders, not with my production data. If you try with a fresh configuration, do you see the black colors like my screenshots ?

My green color is now gone, also the graph segment value colors are now black :-( 

The cotrast is now better :-)

I'll investigate it tomorrow

FJBDev:
> My green color is now gone, also the graph segment value colors are now black :-(
> 
> The cotrast is now better :-)
> 
> I'll investigate it tomorrow

Interesting. Do you think it's a bug ?
I don't have a lot of experience with the tour segmenter as I don't use it a lot so far so i didn't notice.
Let me know what you find

wolfgang-ch:
> 
> 
> > My green color is now gone, also the graph segment value colors are now black :-(
> > The cotrast is now better :-)
> > I'll investigate it tomorrow
> 
> Interesting. Do you think it's a bug ?
> I don't have a lot of experience with the tour segmenter as I don't use it a lot so far so i didn't notice.
> Let me know what you find

It was a bug since ever because the default values were not initialized in the pref store. It was not recognized by me because when you set the graph colors manually (which I did in the beginning) then the colors are set correctly. 

Now the segment default colors are, as they should be :-)

mt-chart-segments-with-correct-default-colors

PR #320, opened 2021-04-06
Map25

telemaxx:
- VTM 0.15
- scaled photos on/off (off is default for reduced resources) (menu item is not localized up to now)


telemaxx:
1st main2main...

PR #321, opened 2021-04-06
Map25: VTM0.15

telemaxx:
- VTM 0.15
- scaled photos on/off (off is default for reduced resources) (menu item is not localized up to now)


telemaxx:
seems so there is a tour import bug.

When i have a fresh MT workspace and imported and saved a tour, tourbook stays empty

wolfgang-ch:
> seems so there is a tour import bug.
> 
> When i have a fresh MT workspace and imported and saved a tour, tourbook stays empty
> 
> Edit: problem of nattable.

NatTable 2.0 should solve the problem

telemaxx:
NatTable 2.0 solved the problem

wolfgang-ch:
Is the released version 0.15 c0d5f4a0cabe7cc6e8ea6cfbf30a1e39e0433984 used?

telemaxx:
no, I cloned the master from vtm. state yesterday. So it's a bit newer.

bundles/net.tourbook/src/net/tourbook/map25/ui/SlideoutMap25_MapOptions.java
```diff
@@ -35,14 +43,6 @@
...
 
-import net.tourbook.Messages;
-import net.tourbook.common.UI;
-import net.tourbook.common.font.MTFont;
-import net.tourbook.common.tooltip.ToolbarSlideout;
-import net.tourbook.map25.Map25App;
-import net.tourbook.map25.Map25ConfigManager;
-import net.tourbook.map25.Map25View;
-
 /**
```
wolfgang-ch, line 46: Blue box is OK, I marked it because I thought it is not working but it is.

Red box: 
* Text key is displayed
* Cannot see a difference when checked or unchecked

mt photo

bundles/net.tourbook/src/net/tourbook/map25/Map25View.java
```diff
@@ -231,22 +232,52 @@
...
+ * private class ActionShowPhoto_WithConfig extends ActionToolbarSlideout {
+ * public ActionShowPhoto_WithConfig() {
+ * super(
+ * TourbookPlugin.getImageDescriptor(Map25View.IMAGE_ACTION_SHOW_PHOTO_IN_MAP),
+ * TourbookPlugin.getImageDescriptor(Map25View.IMAGE_ACTION_SHOW_PHOTO_IN_MAP_DISABLED));
+ * isToggleAction = true;
+ * notSelectedTooltip = MAP_ACTION_SHOW_PHOTO_IN_MAP;
+ * }
+ * @Override
+ * protected ToolbarSlideout createSlideout(final ToolBar toolbar) {
```
wolfgang-ch, line 264: Comments which are outside of a method are moved around by the code formatter, but not always.

wolfgang-ch (commented):
> 
> 
> no, I cloned the master from vtm. state yesterday. So it's a bit newer.

You should write the commit ID which is used into the PR comment, that it could be verified. It is the same as when you get a stacktrace and do no know which version it was used because the code lines could be different.

wolfgang-ch:
> 
> 
> > no, I cloned the master from vtm. state yesterday. So it's a bit newer.
> 
> You should write the commit ID which is used into the PR comment, that it could be verified. It is the same as when you get a stacktrace and do no know which version it was used because the code lines could be different.

I wanted to comment on your answer but now my prepared review comments are also commited, however I was finished with the review.

bundles/net.tourbook/src/net/tourbook/map25/layer/marker/PhotoToolkit.java
```diff
@@ -1,8 +1,8 @@
 /*******************************************************************************
  * Copyright 2016-2018 devemux86
  * Copyright 2017 nebular
- * Copyright 2019, 2020 Wolfgang Schramm and Contributors
- * Copyright 2019, 2020 Thomas Theussing
+ * Copyright 2019, 2020, 2021 Wolfgang Schramm and Contributors
+ * Copyright 2019, 2020, 2021 Thomas Theussing
```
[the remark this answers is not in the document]
telemaxx: i saw that already as i got the conflicting error while merging ;-)

vtm commit
ef2d2dd197e5395b200c6e68a03c3b0805140914

bundles/net.tourbook/src/net/tourbook/map25/ui/SlideoutMap25_MapOptions.java
telemaxx: Wolfgang, did you remember this problem:
https://github.com/wolfgang-ch/mytourbook/commit/d3e09e3afe3c2b6d439ebd9c85afe1603eeb909b#commitcomment-41469693

There you removed the scaling part. but that looks not good on my two W10 laptops and in my VMWare Linux.
Not sure where this depends on.

With scaling on, it looks good everywhere but shows more used resources in "sleak"
with scaling off, it is like in the last version.

grafik

wolfgang-ch: > 
> 
> Wolfgang, did you remember this problem:
> [d3e09e3#commitcomment-41469693](https://github.com/wolfgang-ch/mytourbook/commit/d3e09e3afe3c2b6d439ebd9c85afe1603eeb909b#commitcomment-41469693)
> 
> There you removed the scaling part. but that looks not good on my two W10 laptops and in my VMWare Linux.
> Not sure where this depends on.
> 
> With scaling on, it looks good everywhere but shows more used resources in "sleak"
> with scaling off, it is like in the last version.
> 
> grafik

I tried to understand the photo painting, but giving up for now, why it isn't painted with a larger (not scaled) image.

Can you set the missing text, afterwards I will merge it.

telemaxx: yes of coarse.

should i call it :
"scaled photos"
or better
"force scaled photos"

because sometimes, like in you're case its always scaled.

should i add this line only to messages.properties and Messages.java or also in messages_de.properties?

wolfgang-ch: We should not call it scaled, because scaled could be larger OR smaller.

Also, I think we should look over the "Tellerrand" (don't know if there is an english word for it) and the user should be able to set/select the image size.

There may be fixed sizes, e.g. 160, 300, ... and a variable size.

When you are not yet implementing this sizing feature, then the checkbox label could be "Small photo" which you can also translate into german (very difficult task :-)

telemaxx: the easiest for me is to copy youre hillshading code for maps options

grafik

telemaxx: checkbox and text is not ready yet

telemaxx: like this:
grafik

wolfgang-ch: This is OK

PR #322, opened 2021-04-06
Update required-plugins-in-eclipse-ide.txt

telemaxx:
nattable 2.0.0


telemaxx:
here we should also point to 2.0.0:
https://mytourbook.sourceforge.io/mytourbook/index.php/development/dev-ide/documentation-dev-ide-app-plugins

PR #326, opened 2021-04-10
Map25 resized photos in map25

telemaxx:
with save and restore


telemaxx:
needs some code cleanup, but it works so far.

The conflicts i have solved in the browser, hopefully correctly ;-)

telemaxx:
please dont merge my prefs:
bundles/net.tourbook/.settings/org.eclipse.jdt.ui.prefs

wolfgang-ch (commented):
Resizing of the image works :-)

however only the EXIF thumb image is resized which just enlarges the photo pixels

I merged this PR in main but having this issues:

* I don't see what the checkbox for "Photo Size" is used for
* For what is the checkbox "Slideout_Map25......" used


mt 25map image size

telemaxx:
i have already expected that you're Computer will upscale the thumbnails and shows big pixels...
not sure why my computers have a different behavior.
unscaled i get really big images, which can scaled down with good quality.
That is the reason, why i did not call it "small photos", because in youre case photo becomes bigger.

I thing a better text for the checkbox would be: "Photo resize" or "resize photos". what do you think?

In the 1st version of the photo layer i used the original photos, but that was extremely slow.

> For what is the checkbox "Slideout_Map25......" used
oops, that is a bug. should only the button aside the spinner.
on youre screenshot "Photo Size" Button is unchecked, so it should show unscalled photos.
I will check that also.

wolfgang-ch:
I just adjusted the wording of the `Map 2.5 D Options` labels and commited it, only the first word should be uppercase

wolfgang-ch:
> I thing a better text for the checkbox would be: "Photo resize" or "resize photos". what do you think?

`Photo size` is OK

telemaxx:
on my computer it looks correct. maybe due the conflicts, which i solved via browser.

grafik

telemaxx:
> 
> 
> I just adjusted the wording of the `Map 2.5 D Options` labels and commited it, only the first word should be uppercase

did this not give a conflict when i now send a new PR?

PR #327, opened 2021-04-11
Map25: resending the slideout, which had conflicts

telemaxx:


telemaxx:
i had again the conflicts. so tried eg the flowing deletion via browser:
grafik

can you check if it is removed in this PR?

bundles/net.tourbook/.settings/org.eclipse.jdt.ui.prefs
```diff
@@ -125,15 +159,26 @@ sp_cleanup.remove_trailing_whitespaces_ignore_empty=false
 sp_cleanup.remove_unnecessary_array_creation=false
 sp_cleanup.remove_unnecessary_casts=true
 sp_cleanup.remove_unnecessary_nls_tags=true
-sp_cleanup.remove_unused_imports=true
+sp_cleanup.remove_unused_imports=false
```
wolfgang-ch, line 162: Can you fix this file, this morning I spend time to revert your deletion and now it is modified again. 
This value `sp_cleanup.remove_unused_imports=true` cannot be `false`

All of these files must be the same

mt ui prefs

telemaxx:
now its:
sp_cleanup.remove_unused_imports=true

wolfgang-ch:
> now its:
> sp_cleanup.remove_unused_imports=true

OK, this value is set correctly but you have saved it propably with Eclipse 2021-03 which added a lot of new properties. As I pointed out above, **ALL** `.settings` **MUST** be the same, now they are not.

Can you revert this file back to the original value which I already did this morning.

mt ui prefs

PR #329, opened 2021-04-12
Map25: try to get better image quality also for wolfgangs computer

telemaxx:
from:
final ImageQuality requestedImageQuality = ImageQuality.THUMB;
to
final ImageQuality requestedImageQuality = ImageQuality.HQ;

photo toolkit -> line 304

grafik


telemaxx:
seems to make problems with that big images.

PR #331, opened 2021-04-17
Map25 wording "create Tour Marker"

telemaxx:


telemaxx:
if merged the feature request can be closed:
https://sourceforge.net/p/mytourbook/feature-requests/192/

bundles/net.tourbook/src/net/tourbook/messages_cs.properties
```diff
@@ -3164,7 +3164,7 @@ tag_view_title_tag                   = Zna\u010Dka:
 tag_view_title_tag_category          = Kategorie:
 
 tourCatalog_view_action_create_left_marker           = Vytvo\u0159it zna\u010Dku v&levo
-tourCatalog_view_action_create_marker                = Vytvo\u0159it &zna\u010Dku
+tourCatalog_view_action_create_marker                = Vytvo\u0159it
```
wolfgang-ch, line 3167: @telemaxx Are you doing now the translations :-)

wolfgang-ch:
Still having the issue that the thumb image is scaled

mt map 25 photos

telemaxx:
i know, i had reverted my changes for that.
it had looked good, but when i load a tour with many photos, it crashes.

can you try, what happens, when you change this line:
https://github.com/wolfgang-ch/mytourbook/blob/df98402e4bb417c74301276a260e92439911ea94/bundles/net.tourbook/src/net/tourbook/map25/layer/marker/PhotoToolkit.java#L304

to:

final ImageQuality requestedImageQuality = ImageQuality.HQ;

HQ seems to be to much.

bundles/net.tourbook/src/net/tourbook/messages_cs.properties
telemaxx: was easy, in this case.
just copy past/from the other messages.properties.

but i keep the shortcuts

wolfgang-ch:
When I changed it to the HQ quality, then MT crashed sometimes but the image quality is better :-)


mt map25 photos

I also tried to reduced the cache and photo quality size and it seems to crash less, but still.

I also discovered that photos in the tour photo view are not always loaded with higher quality when the image size is larger than the thumb size (when photos are displaying a small red-orange square in the corner)

mt map25 photos quality


mt map25 photos cache

PR #339, opened 2021-05-14
Map25: direction arrows on the track

telemaxx:
https://sourceforge.net/p/mytourbook/feature-requests/195/

grafik


FJBDev:
Looks like those labels should be vertically aligned
image

telemaxx:
Some slideouts need some rework... when you click the color chooser,  the window opened behind the slideout. @wolfgang-ch solved that behaviour in other slideouts.

telemaxx:
@FJBDev i have shorten the text a bit.
it looks better now but not perfect. not sure how to handle this swt thinks...
grafik

wolfgang-ch:
When I open the color picker dialog also in other slideouts, this dialog is pinned in the center of the MT main window with Ubuntu 20.04, however it can be behind the slideout but it can be move to be fully accessible.

wolfgang-ch:
I did some UI adjustments :-)

image

telemaxx:
thank you,  that's looks good. 
and now the translators can use longer words,  without destroying the layout. 

high tec developer@work:

wolfgang-ch:
I use the same tools (paper and pen) when something gets complicated :-)

and I cannot remember when is Latitude and Longitude, this is my "Eselsbrücke":
It is "Lon" when it starts in "London" which has longitude = 0

telemaxx:
I like this Eselsbrücke :-)

PR #344, opened 2021-05-16
Map25 Adding Tooltips for Slideout. Stepsize for spinner was to fine

telemaxx:
it was not clear to me where the number next to the color was used for. so i added a Tooltip for the opacity.

in the tourtrack slideout two spinner where set to:
`_spinnerOutline_Width.setDigits(1);`
due that `spinnerOutline_Width.setIncrement(1); `  results in a stepsize of 0.1
so if i want to change the spinner value by 2 i have to press the plus key 20 times...
so i set the stepsize to 5.

grafik


wolfgang-ch:
> 
> 
> it was not clear to me where the number next to the color was used for. so i added a Tooltip for the opacity.
> 
> in the tourtrack slideout two spinner where set to:
> `_spinnerOutline_Width.setDigits(1);`
> due that `spinnerOutline_Width.setIncrement(1); ` results in a stepsize of 0.1
> so if i want to change the spinner value by 2 i have to press the plus key 20 times...
> so i set the stepsize to 5.
> 

For all spinner controls, a page up/down is mostly set to 10 which needs less key strokes and also in MT all spinner controls have mouse accelerated buttons, press `<Ctrl>` , `<Shift>`  or `<Ctrl>+<Shift>` when using the mouse wheel

telemaxx:
mouse wheel , I will test. 
I think many users are mousejunkies like me and don't use the accelerators :-)
I have tied different stepsizes.
smaller steps as 0.5 are not visible for track width. I had expected only 1px step would be visible.

telemaxx:
The Textfields of the active Spinners are hard to read.
Compare "Hillshanding" with "photosize"
grafik

Only with the great new dark them.
started from eclipse on linux.

wolfgang-ch:
> grafik
> 
> Only with the great new dark them.
> started from eclipse on linux.

With Ubuntu is looks OK

image

telemaxx:
ah okay. Problem of my very special configuration.

PR #345, opened 2021-05-25
French Translation

FJBDev:
- French translation
- Translation the "Out of Sync" error messages


bundles/net.tourbook.device/src/net/tourbook/device/polar/hrm/messages_fr.properties
```diff
@@ -1,28 +1,71 @@
 #Generated by Eclipse Messages Editor (Eclipse Babel)
 
-Import_Confirm_PPD = 
+Import_Confirm_PPD = Vous \u00EAtes sur le point d'importer un fichier PPD (Polar Person Definition).\n\
```
FJBDev, line 3: @wolfgang-ch I recently found out that the Polar Preference page didn't even have the french translation file.

wolfgang-ch: Translation files are not automatically created with native Eclipse, one of the messages editors may can create the language files as I remember.

bundles/net.tourbook/src/net/tourbook/tour/TourManager.java
```diff
@@ -502,37 +502,15 @@ public static boolean checkTourData(final TourData tourData1, final TourData tou
...
-               + tourData1.toStringWithHash()
-               + UI.NEW_LINE2
-               + UI.NEW_LINE
-               + "You should also inform the author of the application how this error occured. " //                        //$NON-NLS-1$
-               + "However it isn't very easy to find out, what actions are exactly done, before this error occured. " //   //$NON-NLS-1$
-               + UI.NEW_LINE2
-               + "These actions must be reproducable otherwise the bug cannot be identified."; //                          //$NON-NLS-1$
-
-         MessageDialog.openError(Display.getDefault().getActiveShell(), "Error: Out of Synch", message); //                //$NON-NLS-1$
+         MessageDialog.openError(Display.getCurrent().getActiveShell(),
```
wolfgang-ch, line 511: When using `Display.getCurrent()` it can fail because it can be `null` when running not in the UI thread. 
Use `Display.getCurrent()` only when you are sure, the code is running in the UI thread.

bundles/net.tourbook.device/src/net/tourbook/device/polar/hrm/messages_fr.properties
FJBDev: What I meant is that I am surprised nobody noticed/complained that the complete Polar preference page french translation was missing all those years (5+ years ago!)

bundles/net.tourbook/src/net/tourbook/tour/TourManager.java
FJBDev: Fixed.

PR #358, opened 2021-06-07
Using the Job framework in order for the user to continue doing other actions

FJBDev:
@wolfgang-ch @telemaxx 

What do you think of converting, when appropriate only, [IRunnableWithProgress ](http://help.eclipse.org/2021-03/topic/org.eclipse.platform.doc.isv/reference/api/org/eclipse/jface/operation/IRunnableWithProgress.html)to [Job ](http://help.eclipse.org/2021-03/topic/org.eclipse.platform.doc.isv/reference/api/org/eclipse/core/runtime/jobs/Job.html)so that the user can continue doing other actions while a job is performed in the background ?
I had this idea when thinking about how Eclipse handles pushing commits and when uploading the commit, it is a job in the bottom and I can continue coding and am not "blocked"
image

Here is the example with Strava when sometimes I feel that I don't want to watch the upload but only want to know the final result and in the meantime I can continue use MT

Here is the current behavior where the window blocks the user of doing anything else:
image

here is what it looks like after this code change. The progress can be displayed but the user can do other actions and is not "blocked":
image

In both cases, when the operation is done, the summary will be displayed to the user:
image

### DIFF
--- a/bundles/net.tourbook.cloud/src/net/tourbook/cloud/strava/StravaUploader.java
+++ b/bundles/net.tourbook.cloud/src/net/tourbook/cloud/strava/StravaUploader.java
@@ -21,7 +21,6 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
-import java.lang.reflect.InvocationTargetException;
 import java.net.HttpURLConnection;
 import java.net.URI;
 import java.net.http.HttpClient;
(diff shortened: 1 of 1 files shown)


wolfgang-ch:
I think it is a good idea for longrunning tasks to run in the background. You may have to ensure that this task is not started multiple times by the user, when not yet finished.

telemaxx:
cool,  like in eclipse. and not that much code changes. 
not sure what happens, when eg upload is not finished and user delete/rename the tour in Mt.

FJBDev:
> You may have to ensure that this task is not started multiple times by the user, when not yet finished.
You don't think a user could launch the upload of several tours in different tasks ?

>not sure what happens, when eg upload is not finished and user delete/rename the tour in Mt.
Good point, I wonder if a tour can be turned in read-only mode during the upload and brought back to write edit mode after the upload is done ?

wolfgang-ch:
> > You may have to ensure that this task is not started multiple times by the user, when not yet finished.
>>
> You don't think a user could launch the upload of several tours in different tasks ?

 You have to make it bullet proof because of DAUs (DAU=Dümmster Anzunehmender User = most stupid user)

> > not sure what happens, when eg upload is not finished and user delete/rename the tour in Mt.
> > Good point, I wonder if a tour can be turned in read-only mode during the upload and brought back to write edit mode after the upload is done ?

- The tour editor must be in readonly mode
- Any modify actions have to be disabled or the tour save method must display a message box that actions cannot be done and the tour have to be reverted. This can be complicated depending on the performing action.

FJBDev:
>     * Any modify actions have to be disabled or the tour save method must display a message box that actions cannot be done and the tour have to be reverted. This can be complicated depending on the performing action.

It might be more complicated than I originally thought....

wolfgang-ch:
Next GitHub suprise :-(

I deleted fb-improvements branch with the provided delete button, this automatically closes this PR, so the message

image

"... branch can be savely deleted" **is false**

@FJBDev  Can you recreate this PR?

FJBDev:
> Next GitHub suprise :-(
> 
> I deleted fb-improvements branch with the provided delete button, this automatically closed this PR, so the message
> 
> "... branch can be savely deleted" **is false**
> 
> @FJBDev Can you recreate this PR?

@wolfgang-ch 
Again, GitHub is full of surprises!
No worries, this PR was actually not ready since I had not made it bullet proof because of DAUs (DAU=Dümmster Anzunehmender User = most stupid user) 😄 

I will reopen it when it is ready but at the moment I am working on other features.

PR #368, opened 2021-07-15
Fitlog/FitlogEx import: The timezone is used only when available, otherwise it's determined with the GPS position

FJBDev:
Unit tests 100% successful

image


wolfgang-ch:
> 
> 
> Example of my `workbench.xmi`
> 
> ```
> <children xsi:type="advanced:Placeholder" xmi:id="_SjxAqZCdEeuf-4SFt8Xi7g" elementId="net.tourbook.views.rawData.RawDataView" ref="_SjxoPZCdEeuf-4SFt8Xi7g" closeable="true">
>                   <tags>View</tags>
>                   <tags>categoryTag:1. Tour Directories</tags>
>                 </children>
>                 <children xsi:type="advanced:Placeholder" xmi:id="_SjxArZCdEeuf-4SFt8Xi7g" elementId="net.tourbook.views.tourListView" ref="_SjyOwZCdEeuf-4SFt8Xi7g" closeable="true">
>                   <tags>View</tags>
>                   <tags>categoryTag:1. Tour Directories</tags>
>                 </children>
>               </children>
> ```

Both tags are OK but you have to check all `<children xsi:type="advanced:Placeholder"...` tags if it contains the `closeable="true"` attribute

FJBDev:
> Both tags are OK but you have to check all `<children xsi:type="advanced:Placeholder"...` tags if it contains the `closeable="true"` attribute

I checked all of them and they all have the `closeable="true"` attribute

PR #375, opened 2021-08-04
[Map25] direction arrows and activated building layer for online maps

telemaxx:
testing also "draft pull request" from git


telemaxx:
"draft" testing was successful, now set as normal PR

PR #369, opened 2021-07-15
Using concurrency for tour reimport

FJBDev:
- Using multi-threading to speed up the re-import process when having selected a lot of tours (Note: This requires to skip tours for which the file location isn't found)

image

Example of re-importing 64 tours **without** concurrency ⏬ **5mn29s**
image

Example of re-importing 64 tours **with** concurrency ⏬ **Less than 1 minute!!**
image

IMPORTANT: Note that when using concurrency, the "Cancel" button is disabled because it created a lot of issues and exceptions as by the time the user could click on the Cancel button, all the tour re-import tasks had already started (multi-threading).
image

PR #379, opened 2021-08-11
Using concurrency for tour reimport - PR 3

FJBDev:


wolfgang-ch:
I don't see a big performance improvement with the concurrent re-import in this PR, I tested the re-import 4 times with 48 tours which returned these results:


**Current main branch**

   Re-imported in 31.232 s
   Re-imported in 23.963 s
   Re-imported in 23.496 s
   Re-imported in 23.899 s

**Current FJBDev branch**

   Re-imported in 28.019 s
   Re-imported in 22.935 s
   Re-imported in 22.799 s
   Re-imported in 22.703 s


I think the bottleneck are the synched methods. When I implemented the concurrent dbupdate, there was a big performance improvement compared with the none concurrent code.

I know, the tour save method is more complicated and I'm not sure if it is an "easy" task to make it fully concurrent.

FJBDev:
> I don't see a big performance improvement with the concurrent re-import in this PR, I tested the re-import 4 times with 48 tours which returned these results:

Are you reimporting tours that are contained in files that contain only 1 tour ?

As you saw in [my first PR](https://github.com/wolfgang-ch/mytourbook/pull/369), I see a huge improvement probably because all the tours reimported are contained in a file that contains 300+ tours.

> 
> I think the bottleneck are the synched methods. When I implemented the concurrent dbupdate, there was a big performance improvement compared with the none concurrent code.
> 
> I know, the tour save method is more complicated and I'm not sure if it is an "easy" task to make it fully concurrent.

When I tried to make the tour save method fully concurrent, i found that it was not only hard but also risky and created side effects behaviors and errors. Being that it is risky, I chose not to continue researching making it fully concurrent.

If you think that it is not worth the small improvement, I am ok reverting it to non concurrent. Let me know

wolfgang-ch:
> Are you reimporting tours that are contained in files that contain only 1 tour ?

Yes, all of my fit files contain only 1 tour since ever

>If you think that it is not worth the small improvement, I am ok reverting it to non concurrent. Let me know

I think we should keep the current code from the 'main' branch but there are other fixed issues in your PR which we should keep.

Sometime I may try to make the re-import fully concurrent, now I know where are the bottlenecks :-)

FJBDev:
> I think we should keep the current code from the 'main' branch but there are other fixed issues in your PR which we should keep.

That sounds good. I have reverted to the current code from `main `and kept the issue fixes. This is available in this [PR](https://github.com/wolfgang-ch/mytourbook/pull/380)

> Sometime I may try to make the re-import fully concurrent, now I know where are the bottlenecks :-)

I think you might get headaches like I did 😄 ..... but considering you already did the `updateTour `concurrent maybe you'll have more success than me

wolfgang-ch:
For the Battery SoC feature, it needed to reimport some tours which took about 15 minutes this was the reason why I started the concurrent re-import feature now. After some days and lot of reworks of the re-import and tour save code, now it works, fast, very fast and very noisy because of the pc fan :-)

This is the result of 746 tours

```
   Reimporting the following values: • Battery

   1. without concurrency:   Re-imported in 337.423 s
   2. with concurrency:      Re-imported in  55.938 s  ~ 6x faster

```

mt re-import concurrent

The code is still in my `improvements` branch. 

This was the first test and I discovered that all import formats MUST support concurrency because the daum format failed. 

This test was done only for selected .fit files.

FJBDev:
Nice, I look forward to testing it with my FitLogex files.

telemaxx:
amazing!

the area bellow the curve is the same, but now vertical.

wolfgang-ch:
> 
> 
> Nice, I look forward to testing it with my FitLogex files.

I just saw, that all file imports will fail when running concurrently and which are using class fields. 

In the .fit importer, all fields are contained in `net.tourbook.device.garmin.fit.FitData` which is the reason why my first concurrency test succeeded out of the box.

I just did an interface change for the file import, see the comment in the commit https://github.com/wolfgang-ch/mytourbook/commit/53315b355ae156700510773485296e72dced5944

wolfgang-ch:
When logging the previous data, it is slowing down the re-import by about 30%, the 1st run is with logging, the 2nd run is logging only very few events

image

image

wolfgang-ch:
Now all CPU cores are happy

image


The update time for the fulltext index is now also logged

image

wolfgang-ch:
> When logging the previous data, it is slowing down the re-import by about 30%, the 1st run is with logging, the 2nd run is logging only very few events

I changed the logging, when only 1 tour is reimported, then all is logged otherwise only errors are logged

FJBDev:
> I changed the logging, when only 1 tour is reimported, then all is logged otherwise only errors are logged

But I need the logging even if I reimport 100 tours to see what was modified and if the values are correct

wolfgang-ch:
> 
> 
> > I changed the logging, when only 1 tour is reimported, then all is logged otherwise only errors are logged
> 
> But I need the logging even if I reimport 100 tours to see what was modified and if the values are correct

You can now select the option to log details or not

image


When using detailed logging, the performance hog depends on how many values are updated

line 8/23 is without logging -> almost the same time
line 14/28 is with logging -> big time difference 

image

PR #396, opened 2021-09-16
When selecting a tour pause in the Tour Chart, the pause is centered in the 2D Map

FJBDev:
@wolfgang-ch 
That is likely my last PR for 21.9. Once you send the email to the translators, I will do the french translation.


bundles/net.tourbook/src/net/tourbook/ui/tourChart/TourChart.java
```diff
@@ -3947,6 +3963,10 @@ public void removeTourModifySelectionListener(final ITourModifyListener listener
       _tourModifyListener.remove(listener);
    }
 
+   public void removeTourPauseSelectionListener(final ITourPauseSelectionListener listener) {
```
FJBDev, line 3966: @wolfgang-ch I didn't see the equivalent `removeTourMarkerSelectionListener` used anywhere is it not ?

wolfgang-ch: Currently there is no need to call `removeTourMarkerSelectionListener` because this listener is used only in a dialog and when this is closed, the tour chart is also discarded

image

FJBDev: Should it be removed from the code then ??

wolfgang-ch: I think for listener we should keep it, these add/remove methods are created manually and not automatically like set/get methods

bundles/net.tourbook/src/net/tourbook/ui/tourChart/ChartLayerPause.java
```diff
@@ -43,10 +43,9 @@
 
    private ChartLabelPause  _hoveredLabel;
    private ChartLabelPause  _tooltipLabel;
-
    private int              _devXPause;
-   private int              _devYPause;
 
+   private int              _devYPause;
    private TourChart        _tourChart;
```
wolfgang-ch, line 49: This reformatting in your code was propably not intended

image

wolfgang-ch:
> 
> 
> @wolfgang-ch
> That is likely my last PR for 21.9. Once you send the email to the translators, I will do the french translation.

I didn't receive a notification when you modified this comment, I just saw it accidentally

I can start the translation process (after I disabled the map25 view for w10) but your fitlogEx import issue https://github.com/wolfgang-ch/mytourbook/commit/7a9dc1f618ae470c7e2e82b6f17394a25bdae6c4#commitcomment-56631969 should be solved in the final 21.9

bundles/net.tourbook/src/net/tourbook/ui/tourChart/TourChart.java
FJBDev: Ok, I kept those

bundles/net.tourbook/src/net/tourbook/ui/tourChart/ChartLayerPause.java
FJBDev: No I did not. Maybe Eclipse did something without asking me?. Fixed in my [PR](https://github.com/wolfgang-ch/mytourbook/pull/398)

PR #404, opened 2021-09-27
Displaying the previous and new battery start and end voltage values when reimporting battery values. Suunto App Support for multiple accounts

FJBDev:
image

### Suunto App Integration:  Support for multiple accounts

image

**REVERTED BECAUSE OF ISSUES WITH 2.5D & 3D MAP TESTING**

~~Adding the ability to export the 2.5D & 3D map views as images~~

~~Note @telemaxx : I couldn't try this contextual menu on 2.5D map as opening 2.5D map view on my Linux and Eclipse 2021-06 crashes:~~
~~`Inconsistency detected by ld.so: dl-lookup.c: 111: check_match: Assertion `version->filename == NULL || ! _dl_name_match_p (version->filename, map)' failed!`~~


bundles/net.tourbook/src/net/tourbook/importdata/RawDataManager.java
```diff
@@ -2796,16 +2801,15 @@ private void reimportTour_12_RequestFileFromUser(final String inSavedImportFileP
...
-                  importState_Process,
-                  reImportStatus);
-         });
+               inSavedImportFilePathName,
+               outReimportFilePathName,
+               tourData,
+               importState_Process,
+               reImportStatus));
       }
    }
```
wolfgang-ch, line 2814: When the block {} is removed in 1, then it is not easy visible, that the reimport... method is a sub statement. I do a lot of codings and do not know every detail afterwards, so it may be possible, that this the reason why I added the block {}

image

bundles/net.tourbook.common/src/net/tourbook/common/color/Map3GradientColorManager.java
```diff
@@ -904,26 +905,17 @@ private static void readColors() {
...
-
-               colorProfile.setIsActiveColorProfile(true);
-
-               // set only first default color profile as active
-               break;
-            }
-         }
+         // set only first default color profile as active
+         _map3ColorDefinitions.values().forEach(colorDef -> colorDef.getColorProviders().get(0).getMap3ColorProfile().setIsActiveColorProfile(
+               true));
```
wolfgang-ch, line 910: I think the new code will fail with an NPE when `...get(0) == null`, this would not happen with the original code

bundles/net.tourbook/src/net/tourbook/importdata/RawDataManager.java
FJBDev: Reverted

PR #183, opened 2020-05-07
MT 20.next : Cloud Connectivity - Dropbox Integration

FJBDev:
### **Dropbox Integration**
  - Ability to connect a Dropbox account to MyTourbook, and also choosing a specific a folder for this account. This Dropbox folder can be set up as a device to watch. In this case, added or deleted files will be automatically and immediately detected.

Here is the detailed process on how to connect a Dropbox account to MyTourbook:

- Go to the [Preferences](http://mytourbook.sourceforge.net/mytourbook/index.php/documentation/preferences) menu. Select "Cloud Connectivity" and "Dropbox"

<br/>

- Clicking on the "Authorize" button will open a window to log into your Dropbox account (You can create one for free by going to [https://www.dropbox.com](https://www.dropbox.com))

<br/>

- Click on "Allow" to allow your Dropbox account to be connected to your installation of MyTourbook.

<br/>

- You should now receive a successful message mentioning that a Dropbox Access Token Retrieval was retrieved successfully. Also, the token will be displayed in the text entry of the preference page.

<br/>

**IMPORTANT: At this moment, we do not offer manual file import directly from Dropbox as the Dropbox module cannot be referenced directly.
This can easily be circumvented by using the manual import and selecting the dropbox folder on the computer.
To have access to this folder locally, install the Dropbox desktop apps [Windows & Linux](https://www.dropbox.com/install).**
Example with Windows 10
image

- ~~Now, you can import files from your Dropbox account directly into MyTourbook by going to the [Import Page](http://mytourbook.sourceforge.net/mytourbook/index.php/documentation/get-tours/import-tours) and clicking on either the large Dropbox icon or the mini one~~

<br/>

- ~~Clicking on either of those buttons will open a dialog where one or multiple files can be selected to be imported into MyTourbook.~~

<br/>

- ~~Select file(s) and click on "OK" to import them~~

<br/>

- This Dropbox integration offers the capability to be used by the Easy Import.
Open the Import [Configuration Dialog](http://mytourbook.sourceforge.net/mytourbook/index.php/documentation/get-tours/import-tours/import-tours-dashboard#import-config-dialog) and enter "Dropbox" into the Device Folder name

<br/>

- Once you have created an Easy Import configuration and activated the [Easy Import Watcher](http://mytourbook.sourceforge.net/mytourbook/index.php/documentation/get-tours/import-tours/import-tours-dashboard#setup-easy-import), MyTourbook will automatically detect new files to import.

<br/>

[...]

PR #406, opened 2021-10-16
Strava Integration: new features

FJBDev:
- Added the ability to upload a tour with or without its description

- Added the ability to map tour types to Strava activities for a more precise tour upload
  https://sourceforge.net/p/mytourbook/discussion/622811/thread/e07bd4d3d7/

1)  Select the option "Use tour type mapping"
image

2) Restart to activate the Strava Tour Type filters
image

3) You can now assign several tour types to a given Strava activity
In the example below, it means that all the tours of tour type "Hills", "Race", and "Running" will be uploaded as a Running activity in Strava

image

4) WARNING: If the same tour type is mapped to more than 1 strava tour type filter, than an error like below will appear when uploading the tour and the upload will be aborted
image

### DIFF
--- a/bundles/net.tourbook.cloud/src/net/tourbook/cloud/strava/StravaUploader.java
+++ b/bundles/net.tourbook.cloud/src/net/tourbook/cloud/strava/StravaUploader.java
@@ -90,8 +90,6 @@ public class StravaUploader extends TourbookCloudUploader {
 
       super(CLOUD_UPLOADER_ID, Messages.VendorName_Strava);
 
-      _tourExporter.setUseDescription(_prefStore.getBoolean(Preferences.STRAVA_SENDDESCRIPTION));
-
       VelocityService.init();
    }
 


bundles/net.tourbook/src/net/tourbook/ui/views/tourBook/ActionDeleteTour.java
```diff
@@ -313,14 +312,7 @@ public void run() {
 
       if (numSelectedTours < 2) {
 
-         final Runnable deleteRunnable = new Runnable() {
-            @Override
-            public void run() {
-
-               // delete selected tours
```
wolfgang-ch, line 320: This comment is lost

bundles/net.tourbook/src/net/tourbook/ui/views/tourBook/ActionDeleteTour.java
```diff
@@ -313,14 +312,7 @@ public void run() {
...
-               deleteTours(selectedTourIDs, treeSelection[0], selectionForDeletedTours, null);
-            }
-         };
+         final Runnable deleteRunnable = () -> deleteTours(selectedTourIDs, treeSelection[0], selectionForDeletedTours, null);
          BusyIndicator.showWhile(Display.getCurrent(), deleteRunnable);
 
       } else {
```
wolfgang-ch, line 318: The following lines could also be converted into a lambda expression, it's almost the same code

bundles/net.tourbook.cloud/src/net/tourbook/cloud/strava/PrefPageStrava.java
```diff
@@ -238,6 +274,24 @@ public void dispose() {
       super.dispose();
    }
 
+   private void enableControls() {
+
+      final boolean isAuthorized = StringUtils.hasContent(_labelAccessToken_Value.getText()) && StringUtils.hasContent(_labelRefreshToken_Value
+            .getText());
```
wolfgang-ch, line 280: Unlucky auto wrap, when this happens in my code, then I format it this way

image

wolfgang-ch (commented):
This PR is merged with minor issues.

bundles/net.tourbook.cloud/src/net/tourbook/cloud/strava/PrefPageStrava.java
FJBDev: I need to remember to do it manually because Eclipse doesn't seem to be able to do it automatically.
Fixed

bundles/net.tourbook/src/net/tourbook/ui/views/tourBook/ActionDeleteTour.java
FJBDev: Good catch. I put it back.

FJBDev:
> > This PR is merged with minor issues.
> 
> I've not done a functional test of your changes but I think you have tested it

I have tested but I have added screenshot of the tour type mapping functionality so you can give me your opinion on if it's a good solution to address [this feature request](https://sourceforge.net/p/mytourbook/discussion/622811/thread/e07bd4d3d7/)

bundles/net.tourbook.cloud/src/net/tourbook/cloud/strava/PrefPageStrava.java
FJBDev: I do too but I have found that sometimes using the "Clean up" formatting puts it back on 1 line

FJBDev:
> 
> 
> > WARNING: If the same tour type is mapped to more than 1 strava tour type filter, than an error like below will appear when uploading the tour and the upload will be aborted
> 
> Better is to prevent in the tour type filter pref page, that invalid tour type filters are saved
> 
> image

But this feature is a cloud feature and [you told me initially](https://github.com/wolfgang-ch/mytourbook/pull/183#discussion_r421496482) that you don't want any cloud specific code in the non cloud code.

wolfgang-ch:
> 
> 
> > > WARNING: If the same tour type is mapped to more than 1 strava tour type filter, than an error like below will appear when uploading the tour and the upload will be aborted
> > 
> > 
> > Better is to prevent in the tour type filter pref page, that invalid tour type filters are saved
> > image
> 
> But this feature is a cloud feature and [you told me initially](https://github.com/wolfgang-ch/mytourbook/pull/183#discussion_r421496482) that you don't want any cloud specific code in the non cloud code.


When you would try to implement it the "normal" way, you would need in the tour type filter pref page a dependency on the cloud plugin but then you would get a cyclic dependency error, plugin A depends on plugin B and plugin B depends ond plugin A. 

The only solution (which I currently see) is to create a plugin extension point to validate a tour type filter. So it would be possible but it is not the easiest way. When using a plugin extension point then there is no cloud specific code in the non cloud code.

You've already implemented extension points, e.g.  `net.tourbook.extension.upload.CloudUploaderManager.readCloudUploadersExtensions(String)`

wolfgang-ch:
> 
> 
> > A multiline label or textbox, each line contains one tour type filter with it's tour types. A readonly textbox could be used when the user should be able to copy the contained text.
> 
> But that would be redundant information with what is already in the tour type filter pref page

This is just an info that a user can quickly see, if it is mapped and how

wolfgang-ch:
> 
> 
> > When you would try to implement it the "normal" way, you would need in the tour type filter pref page a dependency on the cloud plugin but then you would get a cyclic dependency error, plugin A depends on plugin B and plugin B depends ond plugin A.
> > The only solution (which I currently see) is to create a plugin extension point to validate a tour type filter. So it would be possible but it is not the easiest way. When using a plugin extension point then there is no cloud specific code in the non cloud code.
> > You've already implemented extension points, e.g. `net.tourbook.extension.upload.CloudUploaderManager.readCloudUploadersExtensions(String)`
> 
> That could be a possibility but as you mentioned "but it is not the easiest way", maybe it's not worth it for now since I am only adding 5 tour types for Strava. Maybe when/if other vendors use this feature, it will be worth adding this complexity

It is up to you to go these extra steps or not

PR #414, opened 2021-10-28
Enabling JaCoCo for the code coverage report and adjusting the badges to point to  Wolfgang's repository

FJBDev:
Adding [JaCoCo ](https://www.eclemma.org/jacoco/)to show the code coverage
As of today, my unit tests cover 4.5% of the total code. It's not great but it's a good start 😄 

image

TourData is covered by 39% though

image

Updated the FIT SDK to 21.67


bundles/net.tourbook.cloud/src/net/tourbook/cloud/strava/PrefPageStrava.java
```diff
@@ -266,11 +266,13 @@ private void createUI_30_TourUpload(final Composite parent) {
             GridDataFactory.fillDefaults().applyTo(_chkUseTourTypeMapping);
             _chkUseTourTypeMapping.setText(Messages.PrefPage_UploadConfiguration_Button_UseTourTypeMapping);
             _chkUseTourTypeMapping.setToolTipText(Messages.PrefPage_UploadConfiguration_Button_UseTourTypeMapping_Tooltip);
+            _chkUseTourTypeMapping.addSelectionListener(widgetSelectedAdapter(
+                  selectionEvent -> _linkTourTypeFilters.getControl().setEnabled(_chkUseTourTypeMapping.getSelection())));
          }
```
wolfgang-ch, line 271: The enable/disable method should be called not individually on each control. Mostly the `enable...()` method is used even when there are controls enabled/disabled which are not currently involved in the current action. 

When each control would be individually enabled/disabled in its own selection event, it would be a mess when there are many controls. This philosophy is implemented all over MT

image

wolfgang-ch: When testing the newly created Strava tour type filters without adding tour types to the filter, so the filter was empty, there were several issues in the old MT code which I've fixed with https://github.com/wolfgang-ch/mytourbook/commit/baf99ddd8bd5499abf9be342fd8a9d84df2d6011

FJBDev: 10+ year old!!

FJBDev: > The enable/disable method should be called not individually on each control. Mostly the `enable...()` method is used even when there are controls enabled/disabled which are not currently involved in the current action.
> 
> When each control would be individually enabled/disabled in its own selection event, it would be a mess when there are many controls. This philosophy is implemented all over MT
> 

I usually call `enableControls()` in my code but because the Strava preference page is small and only 1 label needed to be enabled, I called it individually. Fixed in my [PR](https://github.com/wolfgang-ch/mytourbook/pull/415)

pom.xml
```diff
@@ -197,18 +197,41 @@
...
-      </plugins>
+    <plugin>
+	    <groupId>org.eclipse.tycho</groupId>
+	    <artifactId>tycho-surefire-plugin</artifactId>
+	    <version>${tycho.version}</version>
+	    <configuration>
+		    <useUIHarness>true</useUIHarness>
+		    <useUIThread>false</useUIThread>
+		    <product>org.eclipse.platform.ide</product>
+		    <skipTests>true</skipTests>
```
FJBDev, line 208: @wolfgang-ch By default, I have set it so that the unit tests do not run because it takes more time for the build to complete. This way, when you run the build on your machine or for the new MT releases, it takes the same time as before. It is only activated for the builds running on GitHub. But you know how to activate it if you want the unit tests to run

pom.xml
```diff
@@ -197,18 +197,41 @@
...
+		            </goals>
+		    </execution>
+		    <execution>
+			    <id>report</id>
+		            <phase>verify</phase>
+			    <goals>
+				    <goal>report-aggregate</goal>
+		            </goals>
+			    <configuration>
+				    <skip>true</skip>
```
FJBDev, line 228: @wolfgang-ch By default, I have set it so that the code coverage is not run because it takes more time for the build to complete (I ran it on my machine and it took  21 mins instead of 4). This way, when you run the build on your machine or for the new MT releases, it takes the same time as before. It is only activated for the builds running on GitHub.

wolfgang-ch: I think to be more climate friendly, this process should run on github not more than once a day and not after each main commit or each PR

bundles/net.tourbook.cloud/src/net/tourbook/cloud/strava/PrefPageStrava.java
wolfgang-ch: > 
> 
> 10+ year old!!

Created 2007-10-27 488436d17eaa4a7ece3c50519e0c87cb77bc1706

pom.xml
FJBDev: Great idea ! I will set that up. Do you have a preference for what time the daily build should run ?

wolfgang-ch: > 
> 
> Great idea ! I will set that up. Do you have a preference for what time the daily build should run ?

see https://github.com/wolfgang-ch/mytourbook/pull/415#discussion_r739960638

PR #415, opened 2021-10-30
[Strava Pref Page] Using a default listener, [Build] Improvements

FJBDev:


.github/workflows/build.yml
```diff
@@ -1,13 +1,7 @@
...
-  push:
-    branches:
-      - main
-    paths:
-      - '**.java'
-      - '**.yml'
-  pull_request_target:
-    types: [opened, synchronize, reopened]
+  schedule:
+  - cron: "0 0 * * *"
```
FJBDev, line 4: @wolfgang-ch For now, I have set up the build to run every day at 00:00 UTC. Let me know which time you prefer.

wolfgang-ch: > 
> 
> @wolfgang-ch For now, I have set up the build to run every day at 00:00 UTC. Let me know which time you prefer.

I would prefer 6:00 UTC to see also your last modifications

FJBDev: Fixed. But would it include my last modifications ?Are you awake at this time to merge my PRs 😄  ?

wolfgang-ch: In the future you have to go to bed at latest at 6:00 UTC :-)

You could change to a later build time but as I saw in the last months, you did not commit after about 7 am my time

FJBDev: What I meant is more the fact that the build runs on the `main `branch only and I don't commit to your `main `branch. So maybe it should be the time that you go to bed since you only can merge my changes to `main `?

wolfgang-ch: It also runs on other events

image

wolfgang-ch: I've seen the many deletions but did not know what it will do

image

FJBDev: The red part (removed lines) were triggering a build for every push on the main branch when a *.java or *.yml file was committed. It was also triggering a build for every pull request.
This is not the case anymore since I removed those lines

The green part (new lines) means that the build is now "scheduled" to run every day at 6am UTC (on your screenshot it says 0am UTC but in main you can see [here ](https://github.com/wolfgang-ch/mytourbook/blob/172c854d9b5a94536c35497e7771cb2db05df1da/.github/workflows/build.yml#L4)it's 6am UTC).

image

wolfgang-ch: >when a *.java or *.yml file was committed

Do the build now run every day even when there was no commit or PR ?

FJBDev: > Do the build now run every day even when there was no commit or PR ?

Yes. If you look into the Actions tab in your repository, you can see that the last 2 builds were triggered at 6am UTC every day since you merged my [PR](https://github.com/wolfgang-ch/mytourbook/pull/415).

image

image

And you can also see in [SonarCloud](https://sonarcloud.io/project/overview?id=wolfgang-ch_mytourbook) each report for each of those builds

image

wolfgang-ch: OK, tomorrow morning it will also run after my todays updates but then I will pause the main commits to see if it's still running which I think, it should not when there a no main commits.

Each build seems to download "half of the the internet", all Eclipse files, they are cached when you run the build locally.

FJBDev: >OK, tomorrow morning it will also run after my todays updates 

Indeed, and actually i think it will fail since your new commits today have the unit tests failing because of this [commit](https://github.com/wolfgang-ch/mytourbook/commit/99aeed06e24b5807b437fec3ab724e6df5d7534b)

Hint: To fix it, you can either add `@JsonIgnore `on top of the getter or update the unit tests files

image

>I will pause the main commits to see if it's still running which I think, it should not when there a no main commits.

Yes, with the current configuration it should run every day. Do you want me to see if it's possible to run it only if there are new commits?

PR #416, opened 2021-11-03
[GitHub Build] Only run the daily build if there are new commits

FJBDev:
Solution found [here](https://stackoverflow.com/a/67527144/7066681)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,4 +63,4 @@ jobs:
         continue-on-error: true
         name: check latest commit is less than a day
         if: ${{ github.event_name == 'push' }}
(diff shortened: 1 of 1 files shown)


wolfgang-ch:
According to the docu

image

only commits in the main branch should run the build ?

FJBDev:
> only commits in the main branch should run the build ?

With the current build, because I have set it up to be a "scheduled" run, the build will only run on the default (`main`) branch.
[Source](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onschedule): 
_"Scheduled workflows run on the latest commit on the default or base branch"_

PR #427, opened 2021-12-04
[Integration Tests] Adding WorldWeatherOnline tests (#84)

FJBDev:


wolfgang-ch (approved):

wolfgang-ch:
New strange behavior, after your branch is merged with this PR into main, your branch is not connected (in Eclipse) to main as it was before...

image

...even I should delete it according to this info

image

FJBDev:
> New strange behavior, after your branch is merged with this PR into main, your branch is not connected (in Eclipse) to main as it was before...
> 
GitHub is always full of surprise! That's weird.... I've never seen that

> 
> ...even I should delete it according to this info
>
If you want, you can delete it, I can always recreate it for my next PR since I have now the permission to do so 😄

Issue #429, opened 2021-12-14
Problems occurred when invoking code from plug-in: "org.eclipse.jface".


FJBDev:
I see this the exception when selecting a tour with altitude data and a tour with no altitude data when the tour chart is enabled in the view

MyTourbook Version: **21.12.1**

```
java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
	at net.tourbook.algorithm.DouglasPeuckerSimplifier.simplify(DouglasPeuckerSimplifier.java:57)
	at net.tourbook.data.TourData.computeAltitudeUpDown_20_Algorithm_DP(TourData.java:2699)
	at net.tourbook.data.TourData.computeAltitudeUpDown(TourData.java:2646)
	at net.tourbook.ui.views.tourMarker.TourMarkerView$3.update(TourMarkerView.java:659)
	at org.eclipse.jface.viewers.ViewerColumn.refresh(ViewerColumn.java:144)
	at org.eclipse.jface.viewers.AbstractTableViewer.doUpdateItem(AbstractTableViewer.java:396)
	at org.eclipse.jface.viewers.StructuredViewer$UpdateItemSafeRunnable.run(StructuredViewer.java:427)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.jface.util.SafeRunnable.run(SafeRunnable.java:174)
	at org.eclipse.jface.viewers.StructuredViewer.updateItem(StructuredViewer.java:2111)
	at org.eclipse.jface.viewers.AbstractTableViewer.internalRefreshAll(AbstractTableViewer.java:680)
	at org.eclipse.jface.viewers.AbstractTableViewer.internalRefresh(AbstractTableViewer.java:618)
	at org.eclipse.jface.viewers.AbstractTableViewer.internalRefresh(AbstractTableViewer.java:610)
	at org.eclipse.jface.viewers.AbstractTableViewer.lambda$0(AbstractTableViewer.java:572)
	at org.eclipse.jface.viewers.StructuredViewer.preservingSelection(StructuredViewer.java:1400)
	at org.eclipse.jface.viewers.StructuredViewer.preservingSelection(StructuredViewer.java:1361)
	at org.eclipse.jface.viewers.AbstractTableViewer.inputChanged(AbstractTableViewer.java:572)
	at org.eclipse.jface.viewers.ContentViewer.setInput(ContentViewer.java:282)
[...]

PR #430, opened 2021-12-14
#429 Fixed an exception when selecting a tour with altitude data

FJBDev:
tour with no altitude data when the tour chart is enabled in the view

### DIFF
--- a/bundles/net.tourbook/src/net/tourbook/algorithm/DouglasPeuckerSimplifier.java
+++ b/bundles/net.tourbook/src/net/tourbook/algorithm/DouglasPeuckerSimplifier.java
@@ -1,14 +1,14 @@
 /*******************************************************************************
- * Copyright (C) 2005, 2015 Wolfgang Schramm and Contributors
- * 
+ * Copyright (C) 2005, 2021 Wolfgang Schramm and Contributors
+ *
  * This program is free software; you can redistribute it and/or modify it under
  * the terms of the GNU General Public License as published by the Free Software
  * Foundation version 2 of the License.
- * 
+ *
  * This program is distributed in the hope that it will be useful, but WITHOUT
  * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
  * FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
- * 
+ *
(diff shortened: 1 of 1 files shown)


wolfgang-ch:
This time I merged this PR with Eclipse and the PR were closed automatically 👍 

Now I also could delete your branch with the provided button, but I keep it so you can reuse/reset it.

image

wolfgang-ch:
> tour with no altitude data when the tour chart is enabled in the view

I'm wondering why this old algorithm (10+ years) which is heavily used, is still causing problems.

Yesterday I run https://github.com/mergebase/log4j-detector over all my projects on my PC and it did not find the troublesome Log4J 2.x files. Sometimes it is better be stay by using the old versions, 1.x

FJBDev:
> > tour with no altitude data when the tour chart is enabled in the view
> 
> I'm wondering why this old algorithm (10+ years) which is heavily used, is still causing problems.
> 

It was not the first time that I saw this error but yesterday was the first time that I was able to reproduce it.

> Yesterday I run https://github.com/mergebase/log4j-detector over all my projects on my PC and it did not find the troublesome Log4J 2.x files. Sometimes it is better be stay by using the old versions, 1.x

Are you referring to [this issue](https://nvd.nist.gov/vuln/detail/CVE-2021-44228) ?
The only places that MT uses log4j are [here ](https://github.com/mytourbook/mytourbook/blob/main/bundles/net.tourbook.ext.velocity/log4j-1.2.17.jar), [here](https://github.com/mytourbook/mytourbook/blob/main/bundles/net.tourbook.ext.ejb3_3.4.0/slf4j-log4j12-1.5.8.jar) and [here](https://github.com/mytourbook/mytourbook/blob/main/bundles/net.tourbook.ext.ejb3_3.4.0/log4j.jar)
The only places

FJBDev:
> Now I also could delete your branch with the provided button, but I keep it so you can reuse/reset it.
> 
I think I will delete it again as it might help starting with a fresh branch for every PR and maybe reduce the risk of problems.

wolfgang-ch:
> Are you referring to [this issue](https://nvd.nist.gov/vuln/detail/CVE-2021-44228) ?

Yes

PR #439, opened 2022-01-06
Github Build fix

FJBDev:


build/build.net.tourbook.feature/feature.xml
```diff
@@ -35,9 +35,7 @@
          id="feature.net.tourbook.app"
          version="0.0.0"/>
 
-   <includes
-         id="feature.net.tourbook.app.nl"
-         version="0.0.0"/>
+   <includes id="feature.net.tourbook.app.nl" version="0.0.0"/>
```
wolfgang-ch, line 38: You should add a comment why it's formatted this way, otherwise it may be possible that it gets reformatted after a while, like I did it just now https://github.com/mytourbook/mytourbook/commit/efa56be2754c12b4a7c94c68d8b62a98936c30cc

FJBDev: Good idea, fixed

PR #444, opened 2022-01-11
[Tour Export] Bugfix for multiple tour export

FJBDev:


bundles/net.tourbook.common/src/net/tourbook/common/util/FilesUtils.java
```diff
@@ -61,4 +61,13 @@ public static String readFileContentString(final String filePath) {
       }
       return fileContent;
    }
+
+   public static String removeExtensions(String fileName) {
+
+      final int extensionPosition = fileName.indexOf('.');
```
wolfgang-ch, line 67: Instead of `.indexOf` shouldn't it be `.lastIndexOf` as it was in https://github.com/mytourbook/mytourbook/blob/879ffae8bc556cf678870251aa54011d312c5c34/bundles/net.tourbook/src/net/tourbook/srtm/FileZip.java#L44-L46

bundles/net.tourbook.export/src/net/tourbook/export/DialogExportTour.java
```diff
@@ -1443,20 +1469,15 @@ private void setError(final String message) {
...
-            final long tourMillis = tourData.getTourStartTime().toInstant().toEpochMilli();
+         final long tourMillis = tourData.getTourStartTime().toInstant().toEpochMilli();
 
-            if (tourMillis < minTourMillis) {
-               minTourData = tourData;
-            }
+         if (tourMillis < minTourMillis) {
+            minTourData = tourData;
          }
       }
```
wolfgang-ch, line 1482: `minTourMillis` was/is wrong before and after this fix.
`minTourMillis` must be set when/from the value of `minTourData` is set

wolfgang-ch (changes requested):

bundles/net.tourbook.export/src/net/tourbook/export/DialogExportTour.java
FJBDev: Good to know. Fixed

bundles/net.tourbook.common/src/net/tourbook/common/util/FilesUtils.java
FJBDev: Reverted.
But the code is confusing because the `if` uses `.indexOf` but the `substring` uses `lastIndexOf`. Is that normal ?

bundles/net.tourbook.export/src/net/tourbook/export/DialogExportTour.java
```diff
@@ -1470,14 +1470,15 @@ private void setFileName() {
 
       // search for the first tour
       TourData minTourData = _tourDataList.get(0);
-      final long minTourMillis = 0;
+      long minTourMillis = minTourData.getTourStartTime().toInstant().toEpochMilli();
```
wolfgang-ch, line 1473: Are you shure that `.get(0)` is not `null`?

There is also this method `TourData.getTourStartTimeMS()`

bundles/net.tourbook.common/src/net/tourbook/common/util/FilesUtils.java
wolfgang-ch: > But the code is confusing because the if uses .indexOf but the substring uses lastIndexOf. Is that normal ?

This is the original code from Alfred Barten

image

wolfgang-ch: > Reverted

You have not used `.lastIndexOf()` in `removeExtensions()`

bundles/net.tourbook.export/src/net/tourbook/export/DialogExportTour.java
FJBDev: Yes, because the export is not launched if it doesn't contain any tours

See [here](https://github.com/mytourbook/mytourbook/blob/d5982fbcb2c41f1c7437d1a79bc361511498f48d/bundles/net.tourbook/src/net/tourbook/extension/export/ActionExport.java#L77)

bundles/net.tourbook.common/src/net/tourbook/common/util/FilesUtils.java
FJBDev: No, instead I have reverted the `FileZip.java` original file

wolfgang-ch: It works as expected but it is not working as it should. Only the first part (until the first dot) of the filename is returned and not the full filename without extension.

image

image

FJBDev: Changed from `indexOf` to `lastIndexOf`

wolfgang-ch: >it was like this for years

It was just undiscovered, like the log4J issue https://nvd.nist.gov/vuln/detail/CVE-2021-44228 but this bug was a little less critical :-)

FJBDev: Ah ok, I'm glad I discovered a bug in the process 😄

PR #447, opened 2022-01-14
[Map 2D] Added the possibility to show/hide the tour weather icon on the map

FJBDev:
image


wolfgang-ch:
With a dimmed map it's better visible

image

FJBDev:
> With a dimmed map it's better visible
> 
The visibility will depend on the map background

PR #472, opened 2022-02-22
[Calendar View] The combo of years now displays the most recent years at the top as it is more likely to be selected than the oldest years

FJBDev:
**Before (I have to go to the bottom of the combo to select last year's 2021):

image

After (2021 or any recent year is easily selectable) :**

image


wolfgang-ch (approved):
Could not find any issues 👍

PR #475, opened 2022-02-22
[Suunto Integration] Added leading zeros to month, day, hour and minutes in the downloaded file names and fixing the activity type mapping

FJBDev:


bundles/net.tourbook.cloud/src/net/tourbook/cloud/suunto/CustomFileNameBuilder.java
```diff
@@ -39,6 +39,10 @@
    private static IPreferenceStore _prefStore =
          Activator.getDefault().getPreferenceStore();
 
+   private static String addLeadingZero(final int number) {
+      return String.format("%02d", number);
```
wolfgang-ch, line 43: missing `//$NON-NLS-1$`

wolfgang-ch (approved):

PR #515, opened 2022-04-18
Database update refactoring

FJBDev:
- Database Fix for [MTB 22.3.0 missing icon for "Weather-Showers-Scattered"](https://sourceforge.net/p/mytourbook/discussion/622811/thread/11993be696/)

- [Fitlog/FitlogEx Import](https://sourceforge.net/p/mytourbook/discussion/622811/thread/11993be696/): Differentiating the temperature from the device and the temperature manually entered/from a provider
  
- Tour Editor: Fixed a bug where the recorded and paused times were not recomputed when using the feature "Delete Time Slices and Remove Time"


bundles/net.tourbook/src/net/tourbook/database/TourDatabase.java
```diff
@@ -5842,24 +6000,26 @@ private boolean updateDb__2_Data(final Connection conn, final SplashManager spla
...
-         updateDb_027_To_028_DataUpdate(conn, splashManager);
-         updateDb_028_To_029_DataUpdate(conn, splashManager);
-         updateDb_031_To_032_DataUpdate(conn, splashManager);
-         updateDb_033_To_034_DataUpdate(conn, splashManager);
-         updateDb_036_To_037_DataUpdate(conn, splashManager);
-         updateDb_039_To_040_DataUpdate(conn, splashManager);
-         updateDb_041_To_042_DataUpdate(conn);
-         updateDb_042_to_043_DataUpdate(conn, splashManager);
-         updateDb_046_to_047_DataUpdate(conn, splashManager);
+         updateDb_003_To_004_DataUpdate(connection, splashManager);
```
FJBDev, line 6003: @wolfgang-ch 
Note that I haven't refactored all the previous data updates because first, before I spent all this time refactoring those, I want to make sure that the refactoring I did is what you asked.

bundles/net.tourbook/src/net/tourbook/database/TourDatabase.java
```diff
@@ -111,10 +111,11 @@ public class TourDatabase {
     * <li>/net.tourbook.export/format-templates/mt-1.0.vm</li>
     * <li>net.tourbook.device.mt.MT_StAXHandler</li>
     */
-   private static final int TOURBOOK_DB_VERSION = 47;
+   private static final int TOURBOOK_DB_VERSION = 48;
 
-//   private static final int TOURBOOK_DB_VERSION = 48; // 22.X ??
+//   private static final int TOURBOOK_DB_VERSION = 49; // 22.X ??
 
+//   private static final int TOURBOOK_DB_VERSION = 48; // 22.3.1 ??
```
wolfgang-ch, line 118: You should not write a version number `22.3.1 ??` when the version is not yet valid as it is irritating me, better would be `22.x ??`

bundles/net.tourbook/src/net/tourbook/database/TourDatabase.java
```diff
@@ -4866,6 +4867,93 @@ private void modifyColumn_Type(final String table,
...
+    * significantly.
+    *
+    * @param concurrentUpdater
+    *           {@link IComputeNoDataserieValues} interface to concurrently update
+    *           tours
+    * @param tourId
+    *           Tour ID of the tour to be updated
+    * @return
+    */
+   public void performTourDataUpdate_ForAllTours(final long tourId,
```
wolfgang-ch, line 4881: When a method is currently and is propably never used from other files, it should not be public

bundles/net.tourbook/src/net/tourbook/database/TourDatabase.java
```diff
@@ -5506,6 +5594,71 @@ private void sqlStartup_UpgradeDerbyDatabase(final SplashManager splashManager)
       }
    }
 
+   private void updateDatabase(final Connection connection,
```
wolfgang-ch, line 5597: `updateDb__2_Data` calls `updateDatabase` which calls `performTourDataUpdate_ForAllTours`

The code should be kept together by it's naming in an almost 10'000 line file.

I would suggest these naming:

`updateDatabase` -> `updateDb__3_Data_Concurrent`
`performTourDataUpdate_ForAllTours` -> `updateDb__4_Data_Concurrent_OneTour`

Now the methods are all together
image

bundles/net.tourbook/src/net/tourbook/database/TourDatabase.java
wolfgang-ch: I would not refactore an existing very critical code, then you also have to test it very thoroughly **->** so I would also not refactor `TourDataUpdate_046_to_047`

The older I am, the more I do "Don't touch a running system"

wolfgang-ch (commented):
I've not done a funtional test until the issues are fixed

bundles/net.tourbook/src/net/tourbook/database/TourDatabase.java
FJBDev: Fixed

FJBDev:
> I've not done a funtional test until the issues are fixed

That makes sense. It's ready for functional testing as I have addressed your code suggestions/comments

bundles/net.tourbook/src/net/tourbook/database/TourDatabase.java
wolfgang-ch: > > The code should be kept together by it's naming in an almost 10'000 line file.
> 
> That's what makes it hard at times, some files are way too long (`TourData.java`, ....)

With more than 15+ years, a file gets longer but with the orange/green markers on the right side you have a very good overview where a field is read or written or methods are used, I use this feature very often

image

bundles/net.tourbook/src/net/tourbook/database/ITourDataUpdate.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Interface to update tour values for {@link TourData}
+ */
+public interface ITourDataUpdate {
+
+   public int getDatabaseVersion();
+
+   public String getSplashManagerMessage();
+
+   public void updateTourData(TourData tourData);
```
wolfgang-ch, line 29: I think that `updateTourData(...)` should return a boolean, when `true` the tour is saved afterwards, when `false` tour is not saved.

I just saw this improvement when I looked at your `TourDataUpdate_047_to_048.updateTourData` method

wolfgang-ch (commented):
One more improvement

bundles/net.tourbook/src/net/tourbook/database/ITourDataUpdate.java
FJBDev: Good idea, it could make the upgrade faster.

Fixed

bundles/net.tourbook/src/net/tourbook/messages.properties
```diff
@@ -3215,6 +3215,7 @@ Tour_Database_PostUpdate_037_SetHasGeoData        = Data update 37: Set geo data
 Tour_Database_PostUpdate_040_SetTourRecordingTime = Data update 40: Compared tour - {0} of {1}
 Tour_Database_PostUpdate_043_LatLonE6             = Data update 43: Converting lat/lon \u2192 E6 - {0} of {1} - {2} % - {3} \u0394 
 Tour_Database_PostUpdate_047_Weather              = Data update 47: Converting weather data - {0} of {1} - {2} % - {3} \u0394 
+Tour_Database_PostUpdate_048_Weather_Clouds       = Data update 48: Fixing the weather clouds texts - {0} of {1} - {2} % - {3} \u0394
```
wolfgang-ch, line 3218: One minor issue, the text is wrapped and clipped, does this happen also in your dev environment?

image

It seems that this text is larger than a max text length

image

FJBDev: Yes this happens also (english and french). I have fixed it for both languages. However, translators should be let known of that possible issue as it could happen in other languages too.

English
image

French
image

wolfgang-ch: Better...

image

Shouldn't we switch to a common text, e.g.

`Data update %d: %d of %d - %d % - %d Δ`

FJBDev: We could but the user would lose the information about what the update does.

FJBDev: With your suggestion, it now looks like this with my latest PR update

image

wolfgang-ch: > We could but the user would lose the information about what the update does.

We can put this info into the release notes. The info in the splash screen would also not work when there are multiple changes.

wolfgang-ch: > With your suggestion, it now looks like this with my latest PR update
> 
> image

The splash screen looks a bit empty when MT has only 2 tours but now there is enough space for 9.999.999... tours

wolfgang-ch: With more tours it looks less empty

image

bundles/net.tourbook/src/net/tourbook/database/TourDataUpdate.java
```diff
@@ -0,0 +1,45 @@
...
+
+import net.tourbook.data.TourData;
+
+/**
+ * Class to update tour values for {@link TourData}
+ */
+public abstract class TourDataUpdate {
+
+   private int    databaseVersion;
+   private String splashManagerMessage;
```
wolfgang-ch, line 26: After using a common splash message, there is no need to save the splash message in `TourDataUpdate`

The message can be used directly in `TourDatabase`

image


The text key should be adjusted, currently it is `Tour_Database_PostUpdate_048_Weather_Clouds`

As I already suggested, we should not use a formatting mix of NLS and String formatting replacements
`Data update %d: %d of %d - %d % - %d Δ`

bundles/net.tourbook/src/net/tourbook/messages.properties
FJBDev: > > We could but the user would lose the information about what the update does.
> 
> We can put this info into the release notes. The info in the splash screen would also not work when there are multiple changes.

True, it's a good idea to put it in the release notes

bundles/net.tourbook/src/net/tourbook/database/TourDataUpdate.java
FJBDev: Ok, I will do a text key that is generic and remove it from the tour data updater.

FJBDev: Fixed

wolfgang-ch:
Finally, no found issues 👍

PR #881, opened 2022-06-09
22.6.0 fixes & Garmin FIT SDK update

FJBDev:
-  Updating the FIT SDK from 21.78 to 21.84
- English and french typo fixes


wolfgang-ch:
Can you create separate PR's for FitSDK and translations.

It happened one time that a FitSDK had an issue. The current FitSDK is tested very well and the new FitSDK not yet, so I will postpone the FitSDK so MT.next

FJBDev:
> It happened one time that a FitSDK had an issue. The current FitSDK is tested very well and the new FitSDK not yet, so I will postpone the FitSDK so MT.next

If there was an issue, I would think that the FIT unit tests would have caught it. But I understand if you don't want to take any risks.

> Can you create separate PR's for FitSDK and translations.
> 

It's fine, I can wait for the next version, there's only 2 french texts modified so it's not worth all the energy to break this PR into 2 PRs.

wolfgang-ch (approved):

PR #904, opened 2022-08-02
[TCX Export] Fix for the speed that should be in m/s

FJBDev:


wolfgang-ch:
Reviewing a PR in GitHub Browser seems to be limited, I selected the link `Load diff` and received this answer

image

FJBDev:
> Reviewing a PR in GitHub Browser seems to be limited, I selected the link `Load diff` and received this answer

I see the same issue. It seems that GitHub chokes on large files.
I compared it locally with Eclipse/BeyondCompare and it worked.

PR #947, opened 2022-10-16
[SRTM Download] Improved UI when setting tour elevation values from SRTM to display the previous and new elevation gain values

FJBDev:
See original issue [here](https://github.com/mytourbook/mytourbook/issues/334)

image


bundles/net.tourbook/src/net/tourbook/importdata/RawDataManager.java
```diff
@@ -1406,7 +1406,6 @@ private TourData createTourDataDummyClone(final List<TourValueType> tourValueTyp
          for (final TourValueType tourValueType : tourValueTypes) {
 
             final boolean isEntireTour = tourValueType == TourValueType.ENTIRE_TOUR;
-            final boolean isEntireTour_OR_AllTimeSlices = isEntireTour || tourValueType == TourValueType.ALL_TIME_SLICES;
```
wolfgang-ch, line 1409: I don't see a need why this code and the following lines are restructured into 2 methods

FJBDev: I did it because it's hard for me to read and understand [very long and infinite methods](https://github.com/mytourbook/mytourbook/blob/43fa1955d6c21be34644292a929338cc1b0edaf7/bundles/net.tourbook/src/net/tourbook/data/TourData.java#L4623). But I can revert it if you prefer.

FJBDev: > I would like to keep the old code.

Sure, I have reverted the code to the old code

PR #1036, opened 2023-03-23
Unit test fix

FJBDev:


wolfgang-ch:
0 Errors 😄 

image

FJBDev:
> 0 Errors 😄
> 
Weird... I think i happened to me in the past but it was Eclipse that had choked on something.. not sure

PR #1042, opened 2023-03-26
Unit test improvements & 2 specific french translations for 23.3.2

FJBDev:
- Tour Import Unit tests: Stripped down the serialized tours in JSON to represent mostly the most significant tour properties/data that are actually imported from the files into the database and exclude the extraneous data
 
Example: 
- Data serialized in the unit test because it is directly imported from the tour file: pulse, `avgPulse`, distance, `tourDeviceTime_Elapsed` etc....

- Data **NOT** serialized in the unit test because it is **NOT** directly imported from the tour file : distance series converted to double (instead, the  function that converts to double should have its own unit test), data as the weatherPressure etc...

In the future, only the added properties added to `TourData` that are imported directly from a Tour file should be marked with `@JsonProperty` as it ensures that the unit tests verify that this new data is correctly imported into a tour.
That way, when adding a property to TourData that is not coming from a tour file, it will not make the unit tests fail (like in [this case](https://github.com/mytourbook/mytourbook/commit/6f454ed31ab91fde1096f2353f9f7c9bfc5c9db1#commitcomment-106109323)).

- Unit Tests (JUnit Plug-in tests) fixes to succeed on any machine with any given time zone
- 23.3.2 specific translations for two specific texts


wolfgang-ch:
Now with 0 errors and 0 failures 👍 

image

PR #1056, opened 2023-04-04
[Tour values deletion] Added the deletion of the air quality, [Tour Marker View] Added a Pause Type column

FJBDev:
image

- Removed unused messages


bundles/net.tourbook/src/net/tourbook/messages.properties
```diff
@@ -4026,9 +4016,6 @@ tourCatalog_view_action_synch_chart_years_tooltip    = Use the same scaling for
 tourCatalog_view_action_synch_charts_byScale_tooltip = Show the marker in the compared tour with the same scale as in the reference tour
 tourCatalog_view_action_synch_charts_bySize_tooltip  = Show the marker in the compared tour with the same size as in the reference tour
 tourCatalog_view_action_undo_marker_position         = Undo changes, set moved marker to original position
-tourCatalog_view_compare_job_subtask                 = Compared tours: {0} / {1}
-tourCatalog_view_compare_job_task                    = Compare Task
-tourCatalog_view_compare_job_title                   = Compare Tours
```
wolfgang-ch, line 4031: @FJBDev How have you found these text keys, "Find Broken Externalizes Strings" is not working since a while, respectively it finds too many wrong text keys

image

image

FJBDev: > @FJBDev How have you found these text keys, 

I use the [UCDetector plugin](https://marketplace.eclipse.org/content/unnecessary-code-detector) that is very good at finding unused strings, methods etc....

PR #1064, opened 2023-04-14
 [Tour Export] Fit Format & Suunto App Workout upload

FJBDev:
- Tour Export: Added the ability to export tour(s) in FIT format
- Suunto App Integration: Added the ability to upload tours as workouts


bundles/net.tourbook.export/src/net/tourbook/export/fit/FitExporter.java
```diff
@@ -0,0 +1,357 @@
...
+import net.tourbook.common.time.TimeTools;
+import net.tourbook.common.util.StatusUtil;
+import net.tourbook.data.TourData;
+import net.tourbook.data.TourMarker;
+import net.tourbook.data.TourPerson;
+import net.tourbook.export.Activator;
+
+import org.osgi.framework.Version;
+
+public class FitExporter {
```
wolfgang-ch, line 60: Gear ratio is not exported

image

wolfgang-ch: Heartbeat variability is not exported

image

image

FJBDev: I don't have gear ratio so I may have missed that.
I will add gear ratio and HRV to the FIT export

FJBDev: > Gear ratio is not exported
> 
I have fixed this but was not able to thoroughly check since I don't have any files containing gear data

FJBDev: > Heartbeat variability is not exported
> 
I have fixed it and HRV is now exported

wolfgang-ch:
@FJBDev I've sent you 2 private fit files

**2016-06-29-07-57-45.fit**
It do not export gear data that they can be imported

**2019-07-24-06-10-33.fit**
Error: FIT checkFileIntegrity failed 'C:\Users\Wolfgang\Desktop\fit-export-test\2019-07-24_06-10-41.fit' - FIT SDK 21.105

FJBDev:
> @FJBDev I've sent you 2 private fit files
> 
Thank you, I will use them to investigate the issue.

>     * It's also interresting that the fit file name has a difference of a 8 seconds

I notice that too, I'm not sure why since I would think that MT has the same import time than the FIT file ?

FJBDev:
> **2019-07-24-06-10-33.fit**
> 
>     * Garmin Edge 1030
> 
>     * It's also interresting that the fit file name has a difference of a 8 seconds

After analyzing the csv file generated from this FIT file, it seems to me that you started the activity but the actual first timestamp is

Data,2,event,timestamp,"**932875841**",s,timer_trigger,"1",,event,"0",,event_type,"0",,event_group,"0",,,,,

If you convert the timestamp of **932875841** from epoch to human readable time, you obtain 6:10:**41** AM, just like the start time in MT after your import the FIT file

image


And if you dig deeper in the csv file, it seems that in between the 6h10mn33s and 6h10mn41seconds is just a lot of hrv data.

Data,12,hrv,time,"0.566|0.574|65535|65535|65535

So my guess is that the FIT file time (10mn33s) is the time when the device activity started but maybe the actual activity time started (GPS position changed ?) is 10mn41s ?

FJBDev:
> @FJBDev I've sent you 2 private fit files
> 
> **2016-06-29-07-57-45.fit**
> 
>     * Garmin Edge 1000
> 
>     * Exported gear data of this file cannot be imported
> 
> 
> **2019-07-24-06-10-33.fit**
> 
>     * Garmin Edge 1030
> 
>     * Error during import of exported file: FIT checkFileIntegrity failed 'C:\Users\Wolfgang\Desktop\fit-export-test\2019-07-24_06-10-41.fit' - FIT SDK 21.105

@wolfgang-ch  I have fixed the gear issues thanks to your files.

wolfgang-ch:
> So my guess is that the FIT file time (10mn33s) is the time when the device activity started but maybe the actual activity time started (GPS position changed ?) is 10mn41s ?

This is what I also guess, wenn I start a new bicycle tour, I'm standing on my feet with the bicycle between the legs, press the start button, put my left shoe into the click pedal, sitting on the bicycle and then I push the pedal to start. This takes a few seconds until then the autopause recognizes that I'm moving. 

With or without GPS, this doesn't matter because I've a speed sensor when I've not forgotten to enable it.

FJBDev:
>  device data, e.g. Garmin Edge 1000 + Version are also not exported. 
> Exported/Imported
> 
> image

I don't think we should export the original device data (model + version) since the FIT is created from MyTourbook, NOT from the device.  I don't think we should "impersonate" a device from MyTourbook and I would argue that brands like Garmin would frown upon that if they learn that an application is creating FIT files pretending it was created from one of their device.
Additionally, since the device data is [parsed ](https://github.com/mytourbook/mytourbook/blob/60035ec2d53168277283f6609b79de291c86962e/bundles/net.tourbook.device.garmin.fit/src/net/tourbook/device/garmin/fit/FitData.java#L603)when imported in MT, it would not be easy to put it back as it was originally in the original FIT file

wolfgang-ch:
> > device data, e.g. Garmin Edge 1000 + Version are also not exported.
> > Exported/Imported
> > image
> 
> I don't think we should export the original device data (model + version) since the FIT is created from MyTourbook, **NOT** from the device. I don't think we should "impersonate" a device from MyTourbook and I would argue that brands like Garmin would frown upon that if they learn that an application is creating FIT files pretending it was created from one of their device. Additionally, since the device data is [parsed ](https://github.com/mytourbook/mytourbook/blob/60035ec2d53168277283f6609b79de291c86962e/bundles/net.tourbook.device.garmin.fit/src/net/tourbook/device/garmin/fit/FitData.java#L603)when imported in MT, it would not be easy to put it back as it was originally in the original FIT file

This is OK but I would keep the original model + version somewhere in a field or fields or prepend/postpend it, 
e.g. "EDGE 1030" -> "MyTourbook (EDGE 1030)" and similar with the version

FJBDev:
> I agree with you but I would keep the original model + version somewhere in a field or fields or prepend/postpend it, e.g. "EDGE 1030" -> "MyTourbook (EDGE 1030)" and similar with the version

I think if we had a way to know for sure that the tour to be exported was not substantially changed, we could keep the product name but given that we don't know what the user has done, maybe the tour to export is too different from the original one.

Maybe we could put the original model + version in a [Developer field ](https://developer.garmin.com/fit/cookbook/developer-data/#workingwithdeveloperdatafields)?

wolfgang-ch:
> > I agree with you but I would keep the original model + version somewhere in a field or fields or prepend/postpend it, e.g. "EDGE 1030" -> "MyTourbook (EDGE 1030)" and similar with the version
> 
> I think if we had a way to know for sure that the tour to be exported was not substantially changed, we could keep the product name but given that we don't know what the user has done, maybe the tour to export is too different from the original one.
> 
> Maybe we could put the original model + version in a [Developer field ](https://developer.garmin.com/fit/cookbook/developer-data/#workingwithdeveloperdatafields)?

After more thinking about keeping the original model + version, I came to the result to not include it, also because, for the other export formats, we currently do also not include the original model + version values

FJBDev:
> Sensor values and device data, e.g. Garmin Edge 1000 + Version are also not exported. I'll send you another fit file which has all 5 Sensor values

Sensor values export is now fixed
image


> I'm know that sensor values are very tricky to read from the .fit file, so I think that the creation may also be tricky.

yes, it was pretty tricky and took me a while to understand but it should be working now

FJBDev:
> After more thinking about keeping the original model + version, I came to the result to not include it, also because, for the other export formats, we currently do also not include the original model + version values

I agree, that makes sense as it is consistent with the other exports.
This is what is exported in a TCX file [(example)](https://github.com/mytourbook/mytourbook/blob/558063ce098afba1158ae058581943fa54b87e8b/bundles/net.tourbook.tests/src/test/export/files/LongsPeak-HikingActivity.tcx#L10995)
```
<Creator xsi:type="Device_t">
				<Name>MyTourbook 22.6.1.qualifier - https://mytourbook.sourceforge.io</Name>
```

I took the same approach with FIT files ([example](https://github.com/FJBDev/mytourbook/blob/fb-exportfit/bundles/net.tourbook.tests/src/test/export/files/LongsPeakFit.csv)):

`Data,0,device_info,device_index,"0",,manufacturer,"255",,product_name,"MyTourbook",,software_version,"23.3",,timestamp,"962794800",s,`

wolfgang-ch:
> > Sensor values and device data, e.g. Garmin Edge 1000 + Version are also not exported. I'll send you another fit file which has all 5 Sensor values
> 
> Sensor values export is now fixed

👍

wolfgang-ch:
> > I'm know that sensor values are very tricky to read from the .fit file, so I think that the creation may also be tricky.
> 
> yes, it was pretty tricky and took me a while to understand but it should be working now

I used also a feature which was not yet published https://forums.garmin.com/developer/fit-sdk/f/discussion/276245/di2-battery-level/1324422#1324422

bundles/net.tourbook/src/net/tourbook/data/TourData.java
```diff
@@ -7826,7 +7826,7 @@ public short getBattery_Percentage_Start() {
    }
 
    /**
-    * @return Returns battery time in seconds, relative to the tour start time
+    * @return Returns battery time in milliseconds, relative to the tour start time
```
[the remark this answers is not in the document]
FJBDev: >Your comment is correct but the comments of the related fields have still "seconds". I think seconds was intended to be stored but I checked it and milliseconds are stored.

I changed the comment because when I implemented the battery values export, it was confusing because I saw milliseconds for every file so i assumed that it was a bug in the comment.

>This cannot be easily fixed because of the saved tours since it's available. The max value is now roughly 596 hours or 24 days for this battery time field, which is not much but should be OK for 99.9 % use cases.

Maybe you can do a database upgrade to convert from milliseconds to seconds ?

FJBDev:
> I used also a feature which was not yet published https://forums.garmin.com/developer/fit-sdk/f/discussion/276245/di2-battery-level/1324422#1324422

Is it this function that was not yet published at the time (in `RecordMesg`) ?

image

I tried to use it but it didn't work

bundles/net.tourbook/src/net/tourbook/data/TourData.java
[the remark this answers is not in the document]
wolfgang-ch: > > Your comment is correct but the comments of the related fields have still "seconds". I think seconds was intended to be stored but I checked it and milliseconds are stored.
> 
> I changed the comment because when I implemented the battery values export, it was confusing because I saw milliseconds for every file so i assumed that it was a bug in the comment.
> 
> > This cannot be easily fixed because of the saved tours since it's available. The max value is now roughly 596 hours or 24 days for this battery time field, which is not much but should be OK for 99.9 % use cases.
> 
> Maybe you can do a database upgrade to convert from milliseconds to seconds ?

Yes a database upgrade could fix it but it is no "easily" fixed, I would keep it now as it is.

wolfgang-ch:
> > I used also a feature which was not yet published https://forums.garmin.com/developer/fit-sdk/f/discussion/276245/di2-battery-level/1324422#1324422
> 
> Is it this function that was not yet published at the time (in `RecordMesg`) ?
> 
> image image
> 
> I tried to use it but it didn't work

As I remember these `BatterySoc` methods also didn't work 2 years ago, it may work for specific devices. But I requested the Di-2 battery level which is not a simple a "device battery" it's a battery for several devices, front and rear shifters.

PR #1091, opened 2023-05-05
Hibernate 6

FJBDev:
I wanted to learn more about the use of JPA (currently called [EJB ](https://github.com/mytourbook/mytourbook/tree/main/bundles/net.tourbook.ext.ejb3_3.4.0)in the current version of MT) in MyTourbook and decided that updating MyTourbook to a more recent version of Hibernate would be a good exercise since the Hibernate version 3.3.2 was released in [2009](https://github.com/mytourbook/mytourbook/blob/5b94f42e4d0d1bde7996573665fa3fa5041baf67/bundles/net.tourbook.ext.ejb3_3.4.0/lib/hibernate-core-3.3.2/changelog.txt#LL9C41-L9C41)!!
I didn't create this PR to integrate this new Hibernate version in 23.NEXT but this is a draft/prototype PR because I wanted to know your feedback first and that way you can try it locally too.

I didn't notice major differences when using MT with Hibernate 6 (not much faster and not slower) even though I read [several articles](https://jpa-buddy.com/blog/hibernate6-whats-new-and-why-its-important/) that advertised better performances.

I also did a branch where I updated to [Hibernate 5](https://github.com/FJBDev/mytourbook/tree/hibernate5-update). This is the branch I started with in case there would be issues to go from 3.4 to 5. When I saw that it was working fine, I then updated to hibernate6 but kept a separate branch.


wolfgang-ch:
It's good to know that you tested MT with a newer version 😎 

I thought, when there is no need (like using new features) to update EJB then ignore newer versions because it always needs time to do it and you never know if it is better or worser.

> I didn't notice major differences when using MT with Hibernate 6 (not much faster and not slower) even though I read [several articles](https://jpa-buddy.com/blog/hibernate6-whats-new-and-why-its-important/) that advertised better performances.

This was my performance testing, EJB 6 is about 1.3 seconds slower during the app startup. I tested it with an empty workspace which is the fasted startup.

image

[mt-ejb3.log](https://github.com/mytourbook/mytourbook/files/11411522/mt-ejb3.log)
[mt-ejb6.log](https://github.com/mytourbook/mytourbook/files/11411524/mt-ejb6.log)

FJBDev:
> This was my performance testing, EJB 6 is about 1.3 seconds slower during the app startup. I tested it with an empty workspace which is the fasted startup.
> 
> image
> 
Good tests!
I am surprised to see that it's a bit slower but it confirms that it's ok not to upgrade to JPA 6 unless there was bugs.

I am closing this Pull Request but not deleting the branch in case we need to update one day

wolfgang-ch:
I hoped, that newer versions of Hibernate would fix this error message but it is not 😢 

image

FJBDev:
Another reason why i wanted to update to a newer version of Hibernate was to see if it fixes the random issue I have on my production MT in Debian Linux.

But it's hard to know if a new version fixes it because I can't reproduce the issue all the time as it happens randomly

The below error happens even though I don't have MT already running

image

wolfgang-ch:
Are you running the database embedded or with network?

image

FJBDev:
I run "Embedded"

image

FJBDev:
> Is the second error also displayed in your environment?

I will tell you next time it happens (usually in a few days)

wolfgang-ch:
I also have other errors before the SQL exception but the root cause is

`Caused by: ERROR XSDB6: Another instance of Derby may have already booted the database /home/frederic/.mytourbook/derby-database/tourbook.`

wolfgang-ch:
The derby server were never shut down on app closing, which may have caused your issue, this should be now fixed in https://github.com/mytourbook/mytourbook/commit/00bea8f4e527a828fcdeefd50c15e7f5d67a616e

FJBDev:
> The derby server were never shut down on app closing, which may have caused your issue, this should be now fixed in [00bea8f](https://github.com/mytourbook/mytourbook/commit/00bea8f4e527a828fcdeefd50c15e7f5d67a616e)

I have used 23.5 with your fix and so far has not seen this issue again (before, it was happening almost every day).

wolfgang-ch:
I'm happy that this is now fixed, more server shutting down is not possible 😄

FJBDev:
> This was my performance testing, EJB 6 is about 1.3 seconds slower during the app startup. I tested it with an empty workspace which is the fasted startup.
> 
> image
> 

By the way, I assume you used System.currentTimeMillis() to do your measurements ?

Well, if it's the case, it's not a good/accurate way of measuring improvements.
I am watching a Java video and here are the [author](https://github.com/sandermak)'s arguments to NOT do that
Instead, the [Java Microbenchmark Harness JMH ](https://github.com/openjdk/jmh)introduced in Java 12 seems the way to go

image

### Transcript: 
However, please don't write these microbenchmarks in a naive way, yourself. It's very tempting to write code like this where you use System.currentTimeMillis to get the starting time, then some code that you want to benchmark, then user saying call to currentTime release to get the end time, after which you can calculate and print out the elapsed number of milliseconds. This code reads very intuitive and does produce results, but unfortunately, these results aren't very good. **This approach grossly misrepresents what might actually happen during runtime**. **For example, the JVM optimizes code all the time, but it only happens after many runs, and in this approach, you only execute the code to be benchmarked once. The JVM also has all kinds of startup and warmup costs that might mistakenly be included in this approach**. A good benchmark should also do multiple runs and compare the different outcomes to get a realistic picture. **Also, System.currentTimeMillis isn't the most precise way to do timing**. Now, you could address these considerations yourself, but you'll most probably miss some or will get the details wrong. The JMH tool that we're going to look at solves all these issues for you by providing a testing harness that takes care of all the hairy stuff around microbenchmarks. Let's have a look at JMH next.

wolfgang-ch:
The measured time is what the MT startup shows in the console and I assume they do it also with `System.currentTimeMillis()`

To be more precise, I run it mutiple time as I already discovered, that there are time differences between each run, however I think that the overall timer results are correct as it includes the whole MT startup.

When you measure a simple for(...) loop then you can see, that the first runs are slower than the following ones because of the VM optimizations.

FJBDev:
I also wonder if the measurements would be different with Java 17 vs Java 11 since MT is now using Java 17 (since **23.8.0**). Maybe the JDK/JRE has an impact on performance of Hibernate ?

PR #1103, opened 2023-05-12
[23.5] Small fixes per testing

FJBDev:
English and french translation fixes
Removing `:` from labels


bundles/net.tourbook/src/net/tourbook/messages.properties
```diff
@@ -1514,15 +1514,15 @@ Map_Bookmark_Button_Rename                 = &Rename
 Map_Bookmark_Column_Bearing2               = Dir \u00B0
 Map_Bookmark_Column_Bearing2_Tooltip       = Direction
 Map_Bookmark_Column_Latitude               = Lat
-Map_Bookmark_Column_Latitude_Tooltip       = Mapcenter Latitude
+Map_Bookmark_Column_Latitude_Tooltip       = Latitude
 Map_Bookmark_Column_Longitude              = Lon
-Map_Bookmark_Column_Longitude_Tooltip      = Mapcenter Longitude
+Map_Bookmark_Column_Longitude_Tooltip      = Longitude
```
wolfgang-ch, line 1519: Why was `Mapcenter` removed ?

wolfgang-ch (changes requested):

bundles/net.tourbook/src/net/tourbook/messages.properties
FJBDev: Because when i analyzed it, the word center is neither in the variable name nor in the text of other languages so i assumed it was a mistake?

FJBDev: When analyzing, I think there is only german that has the word center in it:
image

If it is meant to have the word center, the variable name `Map_Bookmark_Column_Longitude_Tooltip` and `Map_Bookmark_Column_Latitude_Tooltip` is misleading  but I can revert it to `Mapcenter` (though it should have a space between `Map` and `center`) and update the french translation.

Let me know

wolfgang-ch: `Mapcenter` was also not in the initial implementation

image

But sometimes I changed this wording slightly to better understand this value. It's difficult to find, when I changed this wording without creating new text key's, so I'll do this now after I merged your PR

PR #1115, opened 2023-05-23
Converting the missed beat to heart rate & Geo Compare View fix

FJBDev:
- Geo Compare View:
Fixed a display bug where the maximum opacity value is not displayed

image


bundles/net.tourbook.device/src/net/tourbook/device/messages.properties
```diff
@@ -7,9 +7,9 @@ FitLog_Error_InvalidStartTime       = Tour start time is not set in {0}
 FitLog_Equipment_Name_Not_Available = N/A
 
 Port_Listener_Error_ntd001 = Serial Port Error
-Port_Listener_Error_ntd002 = The serial port could not be opened. \n\
+Port_Listener_Error_ntd002 = The serial port could not be open. \n\
```
wolfgang-ch, line 10: I think that this modification is changing the text wrongly. I've not checked the other `open(ed)` modifications it is just time consuming for me.

I was also not sure if  `R-R ... Beat ...` texts should/must be replaced with `Heart rate` why I didn't changed it in these cases.

image

image

image

wolfgang-ch (changes requested):

PR #1140, opened 2023-06-29
Unit tests fixes

FJBDev:
@wolfgang-ch When some view names are changed, the unit tests need to be updated manually because some of those strings are not in the Java code but only in a [plugin.xml](https://github.com/mytourbook/mytourbook/blob/180a8cb553c4e9c471cb091314e1b8ce2d3a6a47/bundles/net.tourbook/plugin.xml#L600C46-L600C46) file (maybe there is a more dynamic solution that I don't know of?)

### DIFF
--- a/bundles/net.tourbook.ui.tests/src/test/views/TaggedToursViewTests.java
+++ b/bundles/net.tourbook.ui.tests/src/test/views/TaggedToursViewTests.java
@@ -34,7 +34,7 @@ public class TaggedToursViewTests extends UITest {
 
       //Open the Tagged Tours view
       Utils.openOtherMenu(bot);
-      bot.tree().getTreeItem("1. Tour Directories").expand().getNode("Tagged Tours").select(); //$NON-NLS-1$ //$NON-NLS-2$
+      bot.tree().getTreeItem(WorkbenchTests.TOUR_DIRECTORIES).expand().getNode("Tagged Tours").select(); //$NON-NLS-1$
       bot.button("Open").click(); //$NON-NLS-1$
 
       Utils.getTour(bot);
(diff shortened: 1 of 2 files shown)


wolfgang-ch:
> (maybe there is a more dynamic solution that I don't know of?)

There is an answer which is 19 years old 😄 https://www.eclipse.org/forums/index.php/t/76589/

This `org.eclipse.ui.part.WorkbenchPart.setPartName(String)` should do it but should we do it? I've not yet thought enough about it, if it's a good idea or not.

wolfgang-ch:
@FJBDev 

> @wolfgang-ch When some view names are changed, the unit tests need to be updated manually because some of those strings are not in the Java code but only in a [plugin.xml](https://github.com/mytourbook/mytourbook/blob/180a8cb553c4e9c471cb091314e1b8ce2d3a6a47/bundles/net.tourbook/plugin.xml#L600C46-L600C46) file (maybe there is a more dynamic solution that I don't know of?)

With this commit https://github.com/mytourbook/mytourbook/commit/ddfc67ac237c6ece07d68ec31129e2490ad3fac0 you can access the text resources programmatically in `plugin.xml`, see example code in the commit comment

FJBDev:
> With this commit [ddfc67a](https://github.com/mytourbook/mytourbook/commit/ddfc67ac237c6ece07d68ec31129e2490ad3fac0) you can access the text resources programmatically in `plugin.xml`, see example code in the commit comment

It's a good improvement, but it still relies on a string that can change anytime:

`final String viewName = PluginProperties.getText("View_Name_RefTour_ComparedTour");`

wolfgang-ch:
> > With this commit [ddfc67a](https://github.com/mytourbook/mytourbook/commit/ddfc67ac237c6ece07d68ec31129e2490ad3fac0) you can access the text resources programmatically in `plugin.xml`, see example code in the commit comment
> 
> It's a good improvement, but it still relies on a string that can change anytime:
> 
> `final String viewName = PluginProperties.getText("View_Name_RefTour_ComparedTour");`

This solution is not perfect but better than before and text keys 

* should be unique within MT
* can easily be replaced with search/replace opposed to text contents

Issue #1110, opened 2023-05-20
[Feature : Tour Pauses] Add the ability to delete tour pauses


FJBDev:
### What is missing ?

Add the ability to delete tour pauses from the Tour Book View and the Tour Pauses View

wolfgang-ch:
When I delete all pauses in the pauses view, then the pauses are still available in the tour data series. So I don't understand why only one part of the pauses are deleted.

![image](https://github.com/mytourbook/mytourbook/assets/1283445/5e9c81cf-b1c1-49c9-aaad-dcad61ca6966)

FJBDev:
The column you show in your image is a "Break", which is very different than a pause. Deleting a pause won't delete a "Break" as those are computed with an algorithm.
Did you make sure that the pause is not present anymore ?

![image](https://github.com/mytourbook/mytourbook/assets/25235665/97452c2b-287b-4bd8-a7c7-d856933f5d92)

wolfgang-ch:
> Did you make sure that the pause is not present anymore ?

Pauses are gone in the time slice editor


>The column you show in your image is a "Break", which is very different than a pause. Deleting a pause won't delete a "Break" as those are computed with an algorithm.

I was a bit confused but you are correct, pause != break

wolfgang-ch:
When I delete the 3./4. pause, then the 1./2. is removed

![image](https://github.com/mytourbook/mytourbook/assets/1283445/2f393db6-1715-43c3-a354-05b074baa02f)


![image](https://github.com/mytourbook/mytourbook/assets/1283445/f0d2bf39-e8e3-48f6-8bf4-b61b40fde7d2)

FJBDev:
Good catch. I also found a bug when I tested myself this morning.

They should all be fixed in this [PR](https://github.com/mytourbook/mytourbook/pull/1161)

PR #1157, opened 2023-07-31
 [Tour Pauses View] Added the ability to delete tour pauses

FJBDev:
#1110 

image


ovahead:
Great!!

bundles/net.tourbook/src/net/tourbook/ui/views/TourPausesView.java
```diff
@@ -126,19 +130,19 @@ public class TourPausesView extends ViewPart implements ITourProvider, ITourView
...
-      long type;
+      private long type;
 
-      long relativeStartTime;
-      long relativeEndTime;
+      private long relativeStartTime;
+      private long relativeEndTime;
 
-      int  serieIndex;
+      private int  serieIndex;
```
wolfgang-ch, line 140: private fields should start with an `_` https://github.com/mytourbook/mytourbook/discussions/919#discussion-4349284-permalink

bundles/net.tourbook/src/net/tourbook/tour/ActionDeletePausesDialog.java
```diff
@@ -0,0 +1,125 @@
...
+            dialogMessage)) {
+         return;
+      }
+
+      final List<Long> listPausedTime_Start = Arrays.stream(tourData.getPausedTime_Start()).boxed()
+            .collect(Collectors.toList());
+      final List<Long> listPausedTime_End = Arrays.stream(tourData.getPausedTime_End()).boxed()
+            .collect(Collectors.toList());
+      final List<Long> listPausedTime_Data = Arrays.stream(tourData.getPausedTime_Data()).boxed()
+            .collect(Collectors.toList());
```
wolfgang-ch, line 88: I would be easier to read when these 6 lines are not wrapped, line 74 is already longer

wolfgang-ch (commented):
Minor issues

bundles/net.tourbook/src/net/tourbook/tour/ActionDeletePausesDialog.java
FJBDev: Fixed. I typically don't like long line because that requires scrolling in Eclipse but not sure what the code standards are

wolfgang-ch: When I use the default window sizes for my Eclipse IDE and MT windows, then these 3 lines are fully visible 

image

bundles/net.tourbook/src/net/tourbook/ui/views/TourPausesView.java
wolfgang-ch: I'm not fully happy with these private fields. Normally, private field should not be accessible from outside of the class, but they are. So the original implementation with no private fields was more according to the source code rules.

image

FJBDev: This is the [standard practice](https://rules.sonarsource.com/java/RSPEC-1104/) (**Class variable fields should not have public accessibility**) and a rule that makes sense to me.

wolfgang-ch: I don't want to declare all class fields as private and create for each a getter and setter method. I decide, if a field is private or not, depending on the current context.

PR #1160, opened 2023-08-03
 [Tour Pauses Deletion] Changes per code review

FJBDev:
image

Not sure why using ` dialogMessage.append(UI.SYSTEM_NEW_LINE ` makes it look like 2 new lines were added...

### DIFF
--- a/bundles/net.tourbook.export/src/net/tourbook/export/TourExporter.java
+++ b/bundles/net.tourbook.export/src/net/tourbook/export/TourExporter.java
@@ -450,9 +450,7 @@ public class TourExporter {
       if (tourMarkers != null) {
 
          for (final Object element : tourMarkers) {
-            if (element instanceof TourMarker) {
-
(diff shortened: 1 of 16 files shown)


wolfgang-ch:
> Not sure why using `dialogMessage.append(UI.SYSTEM_NEW_LINE` makes it look like 2 new lines were added...

Normally I don't use `UI.SYSTEM_NEW_LINE`, I use `UI.NEW_LINE`

FJBDev:
> Normally I don't use `UI.SYSTEM_NEW_LINE`, I use `UI.NEW_LINE`

I thought it was better to have a [platform-independent new line character](https://stackoverflow.com/questions/207947/how-do-i-get-a-platform-independent-new-line-character)
Then when do you use `UI.SYSTEM_NEW_LINE`?

wolfgang-ch:
I used it only for export features but cannot tell you why

image

PR #1184, opened 2023-09-19
All Tour Markers View renaming and miscellaneous fixes

FJBDev:
- Renamed the "Tour Marker" view to "All Tour Markers" to remove confusion with the "Tour Markers" view
- Fixed typos in Messages
- Using the dynamic way to retrieve the view names
- SonarCloud adjustment for the Java 17 specification


bundles/net.tourbook/src/net/tourbook/tour/DialogSetTimeZone_WizardPage.java
```diff
@@ -15,6 +15,8 @@
  *******************************************************************************/
 package net.tourbook.tour;
 
+import static org.eclipse.swt.events.SelectionListener.widgetSelectedAdapter;
```
wolfgang-ch, line 18: I'm not using such static imports any more because when I change some code in this file and it contains errors and save it, this static import is auto removed and all dependencies in this file are causing errors. This happend to me for several times and since then I don't use these static imports.

You can keep it here for now but when I modify a file and is causing such an error, I remove the static import.

bundles/net.tourbook/plugin.properties
```diff
@@ -200,7 +200,7 @@ View_Name_TourMap25_Tooltip                  = Show tour in 2.5D map
 View_Name_TourMap3                           = 3D Tour Map
 View_Name_TourMap3_Mnemonic                  = 3
 View_Name_TourMap3_Tooltip                   = Show tour in 3D map
-View_Name_TourMarkerAll                      = Tour Marker
+View_Name_TourMarkerAll                      = All Tour Markers
```
wolfgang-ch, line 203: Translators cannot recognize this change, you should delete `View_Name_TourMarkerAll` and use the new text key `View_Name_TourMarkerAll_2`

wolfgang-ch (changes requested):

bundles/net.tourbook/src/net/tourbook/tour/DialogSetTimeZone_WizardPage.java
FJBDev: Interesting, I have never experienced such an error.

I tend to prefer static adapters because they take less lines and are easier to read

bundles/net.tourbook/plugin.properties
FJBDev: Ok, in this case, it might even be better to give a complete new better name as the previous one was not accurate.
Renamed to `View_Name_AllTourMarkers`

bundles/net.tourbook/src/net/tourbook/tour/DialogSetTimeZone_WizardPage.java
wolfgang-ch: > I tend to prefer static adapters because they take less lines and are easier to read

I also would use it but not when I have to correct the code

In the video you can see that when forcing an error and then saving the code, the static import is removed.


https://github.com/mytourbook/mytourbook/assets/1283445/ca2e59fa-4115-4a56-83a0-bd84ee3bf5fa

These are my autosave actions

image

wolfgang-ch (approved):

bundles/net.tourbook/src/net/tourbook/tour/DialogSetTimeZone_WizardPage.java
FJBDev: Interesting, here are my autosave actions but I don't see any obvious difference that would cause that:
But sometimes, Eclipse can be full of surprises that can't be explained

image

PR #1192, opened 2023-10-04
Eclipse 2024-06 with hamcrest build fix and tycho 4 fix

FJBDev:
- Updated to Tycho 4 in order to fix the below issue when updating to Eclipse 2023-09

```
Cannot resolve project dependencies:
[ERROR]   You requested to install 'org.eclipse.equinox.p2.iu; org.hamcrest.core [1.1.0,2.0.0)' but it could not be found
```

- Updated to use Maven 3.9.4 as Tycho 4 required maven >= 3.9.0

- Misc sonarcloud updates


FJBDev:
Setting the PR as draft as it is blocked by this Tycho issue https://github.com/mytourbook/mytourbook/issues/1176

Issue #1195, opened 2023-10-10
[Bug : Tour Blog View] NPE when editing the tour description


FJBDev:
### What happened ?

When clicking on the button to edit the tour description from the Tour Blog View, I get this NPE.

![image](https://github.com/mytourbook/mytourbook/assets/25235665/79f74ac4-73ff-4057-91fc-91df73a729e2)

### Version

23.10

### System

All

### Relevant log output

```shell
java.lang.NullPointerException: Cannot invoke "net.tourbook.data.TourData.getTourTitle()" because "this._tourData" is null
	at net.tourbook.tour.DialogQuickEdit.updateUIFromModel(DialogQuickEdit.java:973)
	at net.tourbook.tour.DialogQuickEdit.createDialogArea(DialogQuickEdit.java:174)
	at org.eclipse.jface.dialogs.TitleAreaDialog.createContents(TitleAreaDialog.java:166)
	at org.eclipse.jface.window.Window.create(Window.java:431)
	at org.eclipse.jface.dialogs.Dialog.create(Dialog.java:1094)
	at net.tourbook.tour.DialogQuickEdit.create(DialogQuickEdit.java:144)
	at org.eclipse.jface.window.Window.open(Window.java:788)
	at net.tourbook.ui.views.TourBlogView.hrefActionEditTour(TourBlogView.java:816)
	at net.tourbook.ui.views.TourBlogView.onBrowserLocationChanging(TourBlogView.java:1005)
	at org.eclipse.swt.browser.LocationListener$1.changing(LocationListener.java:81)
	at org.eclipse.swt.browser.WebKit.webkit_decide_policy(WebKit.java:2304)
	at org.eclipse.swt.browser.WebKit.webViewProc(WebKit.java:616)
	at org.eclipse.swt.browser.WebKit.Proc(WebKit.java:537)
	at org.eclipse.swt.internal.gtk3.GTK3.gtk_main_iteration_do(Native Method)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4514)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1155)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:338)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1046)
[...]

PR #1197, opened 2023-10-11
[Tour Data Editor View] Fix for the NPE when editing the tour's description from the Tour Blog View #1195

FJBDev:


wolfgang-ch:
I could reproduce this issue with your description 👍 

The main issue was, that the tour blog view opened the quick edit dialog when `tourData == null`

When fireing selection events with  `_postSelectionProvider.setSelection` it can have many side effects, so this "solution" should be avoided.

With this commit https://github.com/mytourbook/mytourbook/commit/9a05b07fe4a592b5a609be458f13ff505351dffe the issue in https://github.com/mytourbook/mytourbook/issues/1195 is now fixed

Issue #1193, opened 2023-10-06
[Bug : Tour Editor] Air quality is not displayed when it's saved with a different language


wolfgang-ch:
### What happened ?

The air quality is saved as string with the selected language. The tour editor do not select the saved air quality when the language has changed.

### Version

23.10 and before

### System

All

wolfgang-ch:
I think we have to change this in further versions, that air quality is saved as a number which is independend of the language

Thompson-ongithub:
I don't know if it's the same issue, but already when updating from 23.5 to 23.8 (which I did just recently on my way to test 23.10) whatever is inserted in "air quality" gets removed (MTB in German)

FJBDev:
> I don't know if it's the same issue, but already when updating from 23.5 to 23.8 (which I did just recently on my way to test 23.10) whatever is inserted in "air quality" gets removed (MTB in German)

I can't reproduce the issue on 23.10, can you provide a scenario and/or video capture to show the problem ?

Thompson-ongithub:
In version 23.5 I had a few gps-tracks where "air quality" was set when updating the weather information. After updating this to version 23.8 "air quality" was set to "nicht ausgewählt" (not selected).
After updating the weather information (by clicking on weather, obviously) and getting updates from openweathermap air quality once again was set.

FJBDev:
@Thompson-ongithub When you updated from 23.8 to 23.10, did you have the same issue ? Also, did you happen to switch languages or did you keep using MT in german all that time ?
@wolfgang-ch Since you use german too, did you encounter the issue ?

Thompson-ongithub:
> @Thompson-ongithub When you updated from 23.8 to 23.10, did you have the same issue ? Also, did you happen to switch languages or did you keep using MT in german all that time ? @wolfgang-ch Since you use german too, did you encounter the issue ?

If I remember correctly when upgrading from 23.8 to 23.10 the air quality wasn't removed (I'll double check later this evening)
And no, I never switched langauges. I'm using MTB in German since like forever ;)

Thompson-ongithub:
> > @Thompson-ongithub When you updated from 23.8 to 23.10, did you have the same issue ? Also, did you happen to switch languages or did you keep using MT in german all that time ? @wolfgang-ch Since you use german too, did you encounter the issue ?
> 
[...]

PR #1213, opened 2023-10-19
[Air Quality] Fix and database upgrade to use a unique air quality id instead of a translated text

FJBDev:
#1193


info/release-notes/23.next.0-readme.txt
```diff
@@ -67,7 +67,9 @@ Changes
 Bugfix
 ======
 
-* 
+* Air Quality
```
wolfgang-ch, line 70: The tour info UI displays the saved airquality ID

image

wolfgang-ch (changes requested):

bundles/net.tourbook.common/src/net/tourbook/common/weather/IWeather.java
```diff
@@ -212,6 +218,21 @@ public interface IWeather {
 
    };
 
+   /**
+    * Ids for the weather's air quality. Those Ids are saved in the db.
+    */
+   public static final String[] airQualityIds          = new String[] {
```
wolfgang-ch, line 224: Is SonarCloud not displaying a warning that the name is not UPPERCASE ?

info/release-notes/23.next.0-readme.txt
FJBDev: Good catch, I will fix that

bundles/net.tourbook.common/src/net/tourbook/common/weather/IWeather.java
FJBDev: It would (rule [S116](https://sonarcloud.io/organizations/wolfgang-ch/rules?open=java%3AS116&rule_key=java%3AS116)) but you disabled the rule after [I found out](https://github.com/mytourbook/mytourbook/discussions/919#discussioncomment-3528036) that we couldn't specify a rule for public fields and a different rule for private fields

PR #1239, opened 2023-12-20
[Suunto App] Workout upload translation update, [Slideout Calendar Profiles] Bugfix

FJBDev:
- Package updates and code refactoring
- Slideout Calendar Profiles: Fixed a bug when creating new profiles (Linux only) #1241


bundles/net.tourbook/src/net/tourbook/ui/views/TourBlogView.java
```diff
@@ -454,7 +464,7 @@ private String buildWeatherSection(String tourWeather, final boolean addSpacer)
          tourWeather = UI.scrambleText(tourWeather);
       }
 
-      sb.append("<div class='title'>" + Messages.tour_editor_section_weather + "</div>" + NL); //$NON-NLS-1$ //$NON-NLS-2$
+      sb.append(buildSectionTitle(Messages.Tour_Blog_Section_Weather));
       sb.append("<p class='description'>" + WEB.convertHTML_LineBreaks(tourWeather) + "</p>" + NL); //$NON-NLS-1$ //$NON-NLS-2$
```
wolfgang-ch, line 469: I'm not very happy with this and other refactorings, I would like to see the html tags and it's attributes, e.g. class when I have to look into the html code

image

wolfgang-ch (changes requested):

bundles/net.tourbook/src/net/tourbook/ui/views/TourBlogView.java
FJBDev: >I'm not very happy with this and other refactorings

I reverted to the previous code

PR #1264, opened 2024-01-24
 24.1 Improvement and fix

FJBDev:
- Reducing the large extra space in the Tour Blog View
- Suunto Cloud Downloader: using a copy of the `_prefStore` in the hope of fixing [this issue](https://github.com/mytourbook/mytourbook/pull/1227#issuecomment-1902583767) when launching multiple downloads at a time

### DIFF
--- a/bundles/net.tourbook.cloud/src/net/tourbook/cloud/suunto/SuuntoCloudDownloader.java
+++ b/bundles/net.tourbook.cloud/src/net/tourbook/cloud/suunto/SuuntoCloudDownloader.java
@@ -67,17 +67,17 @@ import org.eclipse.ui.dialogs.PreferencesUtil;
 
 public class SuuntoCloudDownloader extends TourbookCloudDownloader {
 
-   private static final String     LOG_CLOUDACTION_END           = Messages.Log_CloudAction_End;
-   private static final String     LOG_CLOUDACTION_INVALIDTOKENS = Messages.Log_CloudAction_InvalidTokens;
-   private static final String     APP_PEOPLE_ITEM_ALL           = net.tourbook.Messages.App_People_item_all;
+   private static final String LOG_CLOUDACTION_END           = Messages.Log_CloudAction_End;
+   private static final String LOG_CLOUDACTION_INVALIDTOKENS = Messages.Log_CloudAction_InvalidTokens;
+   private static final String APP_PEOPLE_ITEM_ALL           = net.tourbook.Messages.App_People_item_all;
 
-   private static IPreferenceStore _prefStore                    = Activator.getDefault().getPreferenceStore();
-   private int[]                   _numberOfAvailableTours;
+   private IPreferenceStore    _prefStore;
+   private int[]               _numberOfAvailableTours;
 
-   private boolean                 _useActivePerson;
-   private boolean                 _useAllPeople;
+   private boolean             _useActivePerson;
+   private boolean             _useAllPeople;
 
-   private String                  _personName;
+   private String              _personName;
 
    public SuuntoCloudDownloader() {
 
@@ -124,6 +124,8 @@ public class SuuntoCloudDownloader extends TourbookCloudDownloader {
 
    @Override
    public void downloadTours() {
+
+      _prefStore = Activator.getDefault().getPreferenceStore();
 
       _useActivePerson = SuuntoTokensRetrievalHandler.isDownloadReady_ActivePerson();
 
--- a/bundles/net.tourbook.cloud/src/net/tourbook/cloud/suunto/SuuntoCloudDownloader.java
+++ b/bundles/net.tourbook.cloud/src/net/tourbook/cloud/suunto/SuuntoCloudDownloader.java
@@ -148,7 +148,7 @@ public class SuuntoCloudDownloader extends TourbookCloudDownloader {
 
       // Disable the person selector's combo so that the user doesn't launch a concurrent
       // download that could create issues by mixing different user's settings
-      PlatformUI.getWorkbench().getDisplay().asyncExec(() -> PersonManager.get_contributionItem_PersonSelector().setEnabled(false));
+      PlatformUI.getWorkbench().getDisplay().asyncExec(() -> PersonManager.getPersonSelector().setEnabled(false));
 
       final TourPerson activePerson = TourbookPlugin.getActivePerson();
       _personName = activePerson == null
@@ -253,7 +253,7 @@ public class SuuntoCloudDownloader extends TourbookCloudDownloader {
                });
             }
 
-            PlatformUI.getWorkbench().getDisplay().asyncExec(() -> PersonManager.get_contributionItem_PersonSelector().setEnabled(true));
+            PlatformUI.getWorkbench().getDisplay().asyncExec(() -> PersonManager.getPersonSelector().setEnabled(true));
          }
       });
    }
--- a/bundles/net.tourbook.cloud/src/net/tourbook/cloud/suunto/SuuntoCloudDownloader.java
+++ b/bundles/net.tourbook.cloud/src/net/tourbook/cloud/suunto/SuuntoCloudDownloader.java
@@ -49,6 +49,7 @@ import net.tourbook.common.util.SQL;
 import net.tourbook.common.util.StatusUtil;
 import net.tourbook.common.util.StringUtils;
 import net.tourbook.data.TourPerson;
+import net.tourbook.database.PersonManager;
 import net.tourbook.database.TourDatabase;
 import net.tourbook.extension.download.TourbookCloudDownloader;
 import net.tourbook.tour.TourLogManager;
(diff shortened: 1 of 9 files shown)


FJBDev:
@wolfgang-ch 

I tried to disable [this ](https://github.com/mytourbook/mytourbook/blob/main/bundles/net.tourbook/src/net/tourbook/application/ContributionItem_Person.java)`ContributionItem` aka, the combobox in the toolbar but with no success. 
Do you know if it's possible ?

image

wolfgang-ch:
> @wolfgang-ch
> 
> I tried to disable [this ](https://github.com/mytourbook/mytourbook/blob/main/bundles/net.tourbook/src/net/tourbook/application/ContributionItem_Person.java)`ContributionItem` aka, the combobox in the toolbar but with no success. Do you know if it's possible ?
> 
> image

For me it works

image

bundles/net.tourbook.cloud/src/net/tourbook/cloud/suunto/SuuntoCloudDownloader.java
```diff
@@ -125,130 +126,138 @@ private int downloadFiles(final List<Payload> newWorkouts, final IProgressMonito
...
+                           Activator.getImageDescriptor(CloudImages.Cloud_Suunto_Logo_Small),
+                           infoText);
+                  });
+               }
             }
-         }
-      });
+         });
+      }
+      _isInDownload = false;
```
wolfgang-ch, line 260: I think that this flag do not work, this flag should be set to `false` at the end of  the running job and not when the job is scheduled, this returns immediately.

bundles/net.tourbook/src/net/tourbook/ui/views/TourBlogView.java
```diff
@@ -613,8 +606,11 @@ private void create_24_Tour(final StringBuilder sb) {
                                  + " title='" + hoverEdit + "'" //                  //$NON-NLS-1$ //$NON-NLS-2$
                                  + ">" //                                           //$NON-NLS-1$
                                  + "</a>") //                                       //$NON-NLS-1$
-                           + "   </div>" + NL) //                                   //$NON-NLS-1$
-                     + ("<span class='blog-title'>" + tourTitle + "</span>" + NL)); //$NON-NLS-1$ //$NON-NLS-2$
+                           + "   </div>" + NL)); //                                   //$NON-NLS-1$
+
+               if (net.tourbook.common.util.StringUtils.hasContent(tourTitle)) {
```
wolfgang-ch, line 611: Can you use the flag `isTitle` instead of `net.tourbook.common.util.StringUtils.hasContent(tourTitle)` which already checks the content

wolfgang-ch (changes requested):

bundles/net.tourbook.cloud/src/net/tourbook/cloud/suunto/SuuntoCloudDownloader.java
wolfgang-ch: `_isInDownload` may needs to be `static` it may also needs to be `final` then the trick is to use a `boolean[]` with one element

bundles/net.tourbook/src/net/tourbook/ui/views/TourBlogView.java
FJBDev: Good idea, fixed

FJBDev:
> For me it works
> 
> image

For me too it works when I try from inside the class. But my issue is how to access the combo box from an external class like [SuuntoCloudDownloader](https://github.com/mytourbook/mytourbook/blob/main/bundles/net.tourbook.cloud/src/net/tourbook/cloud/suunto/SuuntoCloudDownloader.java)

bundles/net.tourbook.cloud/src/net/tourbook/cloud/suunto/SuuntoCloudDownloader.java
FJBDev: Based on this, I have changed the fix to use a property of `_prefStore` instead of static. That way, if a new download is launched the current `_prefStore` is unchanged.

wolfgang-ch: I don't unterstand this new code, the reference of a pref store is static and I don't see a difference, when you apply the static pref store to a field, it returns always the same static pref store.

In `TourbookPlugin` I already simplified it

image

wolfgang-ch: I think the approach with `_isInDownload` was OK but it was reseted at the wrong time/location. When this would be a very critical concurrent case, then an AtomicBoolean could be used.

wolfgang-ch:
> > For me it works
> 
> For me too it works when I try from inside the class. But my issue is how to access the combo box from an external class like [SuuntoCloudDownloader](https://github.com/mytourbook/mytourbook/blob/main/bundles/net.tourbook.cloud/src/net/tourbook/cloud/suunto/SuuntoCloudDownloader.java)

When the combo is created, store a reference of it in `PersonManager` from where you can get it.

bundles/net.tourbook.cloud/src/net/tourbook/cloud/suunto/SuuntoCloudDownloader.java
FJBDev: That's a possibility but for the moment i think i prefer deactivating the person selector

FJBDev:
> When the combo is created, store a reference of it in `PersonManager` from where you can get it.

Thanks for the tip. I've implemented that and it works now

bundles/net.tourbook.cloud/src/net/tourbook/cloud/suunto/SuuntoCloudDownloader.java
FJBDev: Actually, even using the boolean approach would not work because the issue here is that while the first download runs, if the preference store value changes, then the current download has erroneous results. Hence the fix to disable the person combo is the best so far.

>I don't unterstand this new code, the reference of a pref store is static and I don't see a difference, when you apply the static pref store to a field, it returns always the same static pref store.

My idea was to get a copy of the preference store that would be a copy, i.e.: it would not change when the user does actions during the download but evidently, it doesn't seem to be possible

bundles/net.tourbook.cloud/src/net/tourbook/cloud/suunto/SuuntoCloudDownloader.java
```diff
@@ -247,6 +252,8 @@ public void done(final IJobChangeEvent event) {
                         infoText);
                });
             }
+
+            PlatformUI.getWorkbench().getDisplay().asyncExec(() -> PersonManager.get_contributionItem_PersonSelector().setEnabled(true));
```
wolfgang-ch, line 256: I think that the enablement should be put into a `try {} finally {}` block to ensure it is enabled when an exception occurs

bundles/net.tourbook/src/net/tourbook/database/PersonManager.java
```diff
@@ -29,117 +30,126 @@
...
-	private static final Object				LOCK	= new Object();
+   private static final Object            LOCK    = new Object();
 
-	public static PersonManager getInstance() {
+   private static ContributionItem_Person _contributionItem_PersonSelector;
 
-		if (_instance != null) {
-			return _instance;
-		}
+   public static ContributionItem_Person get_contributionItem_PersonSelector() {
```
wolfgang-ch, line 41: Can you rename the `get/set` methods to the same name as in `net.tourbook.application.ApplicationActionBarAdvisor.getPersonSelector()`, remove `_contributionItem_`

bundles/net.tourbook/src/net/tourbook/ui/views/TourBlogView.java
```diff
@@ -580,23 +580,16 @@ private void create_24_Tour(final StringBuilder sb) {
...
-                  tourTitle = "&nbsp;"; //$NON-NLS-1$
-
-               } else {
-
-                  if (UI.IS_SCRAMBLE_DATA) {
-                     tourTitle = UI.scrambleText(tourTitle);
-                  }
+               if (UI.IS_SCRAMBLE_DATA) {
+                  tourTitle = UI.scrambleText(tourTitle);
                }
```
wolfgang-ch, line 593: This part could be move to the location in the code, where the title is applied

wolfgang-ch (changes requested):

bundles/net.tourbook/src/net/tourbook/database/PersonManager.java
FJBDev: Fixed

bundles/net.tourbook.cloud/src/net/tourbook/cloud/suunto/SuuntoCloudDownloader.java
FJBDev: Fixed

bundles/net.tourbook/src/net/tourbook/ui/views/TourBlogView.java
FJBDev: You mean on the line below ?
`final String hoverEdit = WEB.escapeSingleQuote(NLS.bind(Messages.Tour_Blog_Action_EditTour_Tooltip, tourTitle));`

wolfgang-ch: I thought it could be moved into `if (isTitle...` but the title was already used before, so ignore my previous comment

bundles/net.tourbook.cloud/src/net/tourbook/cloud/suunto/SuuntoCloudDownloader.java
wolfgang-ch: I removed the `try {} finally {}` because I discovered, that the `done()` method in the job listener is always called also by failures

image

FJBDev: Correct, it is the final code called at the end of the Job

PR #1288, opened 2024-02-07
Misc changes and minor fixes

FJBDev:
- Adding the new 3D logo to the README.md file
- Replacing the http urls with **https** for more security


bundles/net.tourbook/src/net/tourbook/tour/TourInfoUI.java
```diff
@@ -15,6 +15,9 @@
  *******************************************************************************/
 package net.tourbook.tour;
 
+import static org.eclipse.swt.events.ControlListener.controlResizedAdapter;
+import static org.eclipse.swt.events.SelectionListener.widgetSelectedAdapter;
```
wolfgang-ch, line 19: This was intentionally **NOT** used because when there is an error in this file, these imports are removed by Eclipse and all related imports are broken -> even more errors in this file

bundles/net.tourbook/src/net/tourbook/ui/views/TourBlogView.java
```diff
@@ -205,32 +204,7 @@ protected ToolbarSlideout createSlideout(final ToolBar toolbar) {
...
-         @Override
-         public void partInputChanged(final IWorkbenchPartReference partRef) {}
-
-         @Override
-         public void partOpened(final IWorkbenchPartReference partRef) {}
-
-         @Override
-         public void partVisible(final IWorkbenchPartReference partRef) {}
-      };
+      _partListener = new IPartListener2() {};
```
wolfgang-ch, line 207: When the part listener is empty, then there is no need to set a part listener at all -> remove `addPartListner`

wolfgang-ch (changes requested):

bundles/net.tourbook/src/net/tourbook/tour/TourInfoUI.java
FJBDev: I've never had issues with those but I have reverted them.

bundles/net.tourbook/src/net/tourbook/ui/views/TourBlogView.java
FJBDev: Ok, good to know. Any reason why it was here in the first place ? It's confusing to see something in the code if it's not used. 
Removed

wolfgang-ch: > Any reason why it was here in the first place ?

I propably copied an existing view and there was no need to use the part listener

bundles/net.tourbook/src/net/tourbook/tour/TourInfoUI.java
wolfgang-ch: I cannot remember if this happened since the beginning but since I found when this issue happens, I don't use the static imports for these listerners


https://github.com/mytourbook/mytourbook/assets/1283445/914f1aa7-a250-487b-bce3-7ca18f837bcf

bundles/net.tourbook/src/net/tourbook/ui/views/TourBlogView.java
FJBDev: That would make sense. I removed all of them to remove confusion

bundles/net.tourbook/src/net/tourbook/tour/TourInfoUI.java
FJBDev: I think it's an Eclipse configuration issue on your Eclipse because I've never had this problem

Uploading Untitled Project.mp4…

wolfgang-ch: I already searched in the Eclipse save actions but have not found the reason why this happens

Issue #1309, opened 2024-03-06
[Bug : Tour Nutrition] The beverage icon is not horizontal adjusted when column is resized


wolfgang-ch:
The beverage icon is not horizontal adjusted when column is resized

https://github.com/mytourbook/mytourbook/assets/1283445/2434c33c-9cb4-463d-9bbe-2cfa144f5466

In the tourbook view the icon is horizontally aligned when resized

https://github.com/mytourbook/mytourbook/assets/1283445/1dac677e-edac-4ac8-b0a4-bc7d3b54b24a

or

https://github.com/mytourbook/mytourbook/assets/1283445/32e7f199-e056-4f44-b787-ff0d21a136a4

_Originally posted by @wolfgang-ch in https://github.com/mytourbook/mytourbook/discussions/1308#discussioncomment-8689532_

PR #1311, opened 2024-03-07
[Tour Nutrition] The beverage icon is now horizontal adjusted when column is resized

FJBDev:
#1309


wolfgang-ch (approved):
I saw in the commit history, that this was not a very simple task but now it works 👍

FJBDev:
> I saw in the commit history, that this was not a very simple task but now it works 👍

It would be too easy if it needed only 1 commit 😄

PR #1326, opened 2024-04-23
[Weather] Migrating to OpenWeatherMap 3.0

FJBDev:
Per the official OpenWeatherMap [announcement](https://openweathermap.org/api/one-call-api), the v2.5 API will be deprecated in June 2024. 
Following this announcement, MT now uses the v3 of the API. The benefits is that the weather retrieval is not limited to the previous 5 days anymore but can be retrieved for **the last 40+ years (from 1st January 1979)**

The weather retrieval remains **free** <ins>but is limited to 1000 requests per day for all MT users.</ins>

`One Call API 2.5 will be deprecated in June 2024. Please read the [detailed guide](https://openweathermap.org/one-call-transfer) for transitioning to the more advanced [One Call API 3.0](https://openweathermap.org/api/one-call-3) which includes several useful upgrades like 40+ years of history archive, daily aggregation and human readable daily summary in comparison with One Call API 2.5 version. Read more about One Call API 3.0 in the product documentation [page](https://openweathermap.org/api/one-call-3) and[ FAQ](https://openweathermap.org/faq#onecall).`

- Tour Nutrition: Support for fractions of servings, disposal of the Dialog Search Product image


bundles/net.tourbook/src/net/tourbook/messages.properties
```diff
@@ -2333,9 +2333,13 @@ Pref_Weather_Check_SaveLogInTourWeatherDescription_Tooltip = When checked, all t
...
-Pref_Weather_Label_OpenWeatherMap_FiveDaysLimit            = NOTE:\n\
-                                                             While the weather retrieval with this vendor is free,\n\
-                                                             historical weather data can only be retrieved for the previous 5 days.
+Pref_Weather_Label_OpenWeatherMap_TimeRangeLimit           = NOTE:\n\
+                                                             While the weather retrieval for this vendor is free and \n\
+                                                             the historical weather data can be retrieved for the last \n\
+                                                             40+ years (from 1st January 1979), the total number of \n\
+                                                             requests per day is limited to 1000. Please be mindful of \n\
+                                                             this as this total number of daily requests is shared across \n\
+                                                             all the users.
```
wolfgang-ch, line 2342: The text in text/label controls should always grab the whole available horizontal space...

image

...like in the import page...

image

...only the tooltip text should be wrapped manually

FJBDev: Fixed

image

wolfgang-ch: After thinking more about the request max limit, we should limit the request per day/session because with just a few clicks in MT, you can easily intentionally or accidentally reach the max. limit.

image

The wording "Retrieve Weather Data..." is irritating because "..." means that there will be a dialogbox to setup/confirm this action but in this case, the weather retrieval starts immediately

FJBDev: > After thinking more about the request max limit, we should limit the request per day/session because with just a few clicks in MT, you can easily intentionally or accidentally reach the max. limit.

That is an idea yes. What do you think the limit should be per user per day ?

What mechanism do you have in mind in order to save the number of requests made per day, using the state like [this one](https://github.com/FJBDev/mytourbook/blob/c14dfd2c98a4f8033e40958d3c2545bdcae55e5b/bundles/net.tourbook/src/net/tourbook/ui/views/nutrition/TourNutritionView.java#L143)?
> 

> 
> The wording "Retrieve Weather Data..." is irritating because "..." means that there will be a dialogbox to setup/confirm this action but in this case, the weather retrieval starts immediately

It takes more to irritate me 😄  but I can change it if you want, to `Retrieve Weather Data` without the `...`

wolfgang-ch: > > After thinking more about the request max limit, we should limit the request per day/session because with just a few clicks in MT, you can easily intentionally or accidentally reach the max. limit.
> 
> That is an idea yes. What do you think the limit should be per user per day ?

20 ... 40 ? 

I think this would be enough when a few tour days are not imported and is mostly more than the current limit of 5 days

wolfgang-ch: > > After thinking more about the request max limit, we should limit the request per day/session because with just a few clicks in MT, you can easily intentionally or accidentally reach the max. limit.
> 
> What mechanism do you have in mind in order to save the number of requests made per day, using the state like [this one](https://github.com/FJBDev/mytourbook/blob/c14dfd2c98a4f8033e40958d3c2545bdcae55e5b/bundles/net.tourbook/src/net/tourbook/ui/views/nutrition/TourNutritionView.java#L143)?

When not necessary, I prefer the state. In the beginning of MT, I only used the pref store because I was not aware of the states, both have advantages/disadvantages

FJBDev: > After thinking more about the request max limit, we should limit the request per day/session because with just a few clicks in MT, you can easily intentionally or accidentally reach the max. limit.
> 

Suggestion implemented. I have decided to limit each user to 50 requests per day. I chose 50 requests because for a given tour, a request will be made for each hour of the tour and therefore retrieving data for several tours can quickly request multiple requests

image


Preference message updated to show to the user the 50 max requests per day
image

bundles/net.tourbook/src/net/tourbook/weather/openweathermap/OpenWeatherMapRetriever.java
```diff
@@ -268,14 +325,27 @@ private TimeMachineResult deserializeWeatherData(final String weatherDataRespons
...
+   private int getRequestCount() {
+
+      int requestCount;
+      try {
+
+         requestCount = _state.getInt(STATE_REQUEST_COUNT);
+
+      } catch (final NumberFormatException e) {
+         requestCount = 0;
+      }
```
wolfgang-ch, line 345: For getting state values you should use the `net.tourbook.common.util.Util.getState...()` methods

bundles/net.tourbook/src/net/tourbook/weather/openweathermap/OpenWeatherMapRetriever.java
```diff
@@ -226,6 +240,24 @@ private String buildWeatherApiRequest(final long date) {
       return weatherRequestWithParameters.toString();
    }
 
+   private boolean canMakeRequest() {
+
+      final String lastRequestDayString = _state.get(STATE_LAST_REQUEST_DAY);
+      final LocalDate lastRequestDay = lastRequestDayString != null ? LocalDate.parse(lastRequestDayString) : LocalDate.of(2000, 1, 1);
+      int requestCount = getRequestCount();
```
wolfgang-ch, line 247: You should use the `net.tourbook.common.util.Util.getState...()` methods, this is also available for date values

wolfgang-ch (changes requested):

bundles/net.tourbook/src/net/tourbook/weather/openweathermap/OpenWeatherMapRetriever.java
```diff
@@ -394,23 +482,31 @@ private boolean retrieveHistoricalWeather() {
    @Override
    public boolean retrieveHistoricalWeatherData() {
 
+      // If the user has reached the maximum number of requests per day, we notify
+      // them and abort the retrieval
+      if (!canMakeRequest()) {
+
+         TourLogManager.log_ERROR(Messages.Log_HistoricalWeatherRetriever_003_RetrievalLimitReached);
+         return false;
+      }
```
wolfgang-ch, line 491: Can you stop the retrieval logging after reaching the limit, othewise the log is filled?

image

bundles/net.tourbook/src/net/tourbook/weather/openweathermap/OpenWeatherMapRetriever.java
FJBDev: Good to know, those functions are useful. 
Fixed

bundles/net.tourbook/src/net/tourbook/weather/openweathermap/OpenWeatherMapRetriever.java
FJBDev: >  this is also available for date values

It is but t[he function](https://github.com/mytourbook/mytourbook/blob/878eef4605f5bc6632430dd8431cae2e3811376d/bundles/net.tourbook.common/src/net/tourbook/common/util/Util.java#L1154) needs a `dateTimeControl`which I don't have in this case. Instead, I used the `getStateString` function

bundles/net.tourbook/src/net/tourbook/weather/openweathermap/OpenWeatherMapRetriever.java
FJBDev: Good idea
Fixed

Example: Below, I have selected several tours but only 1 error message is displayed

image

bundles/net.tourbook/src/net/tourbook/weather/openweathermap/OpenWeatherMapRetriever.java
wolfgang-ch: Why not extend these tools, e.g.

```
public static LocalDate getStateDate(final IDialogSettings state,
                                        final String stateKey,
                                        final LocalDate defaultValue) {

      final String value = state.get(stateKey);

      try {

         return LocalDate.parse(value);

      } catch (final Exception e) {

         return defaultValue;
      }
   }

   public static LocalDate getStateDate(final IDialogSettings state,
                                        final String stateKey,
                                        final LocalDate defaultValue,
                                        final DateTime dateTimeControl) {

      final LocalDate parsedValue = getStateDate(state, stateKey, defaultValue);

      dateTimeControl.setYear(parsedValue.getYear());
      dateTimeControl.setMonth(parsedValue.getMonthValue() - 1);
      dateTimeControl.setDay(parsedValue.getDayOfMonth());

      return parsedValue;
   }
```

FJBDev: >Why not extend these tools, e.g.

Usually, I extend tools whenever a function is used at several places and not only once.

But since you're asking, I have added the function per your suggestion

PR #1354, opened 2024-05-25
improve the Spanish translation that was missing

danielrtic:


wolfgang-ch:
Thanks for the spanish translation

PR #1418, opened 2024-10-23
24.11 French translation

FJBDev:


bundles/net.tourbook.chart/src/net/tourbook/chart/ChartDrawingData.java
```diff
@@ -27,7 +27,7 @@ public class ChartDrawingData {
    ChartDataModel                      chartDataModel;
 
    int                                 devMarginTop;
-   int                                 devXTitelBarHeight;
+   int                                 devXTitleBarHeight;
    int                                 devSliderBarHeight;
```
wolfgang-ch, line 31: This issue exists since the beginning

image

PR #1478, opened 2025-04-18
 Tour Nutrition Enhancements

FJBDev:
* Tour Info Tooltip:
  - New fields: elapsed time at night and during the day for tours that also happen during the night

image

* Tour Nutrition:
  - Support for the carbohydrates
  - New button to update the nutrition facts of (a) product(s)
image


wolfgang-ch:
It is great that you can continue your contributions to MT.

I've not yet looked into the code but can the new fields in the tooltip be optional like other fields?

I'm waiting with this merge until MT 25.4 is published and this will be after eastern.

The last few days I did not contributed any to MT because now I need time to use it by sorting out my 14 year old photos during my Yellowstone tour. Just yesterday I created these images

2011 USA - 30 Yellowstone - 1  Grizzly

2011 USA - 32 Yellowstone - Fuchsbau

FJBDev:
> It is great that you can continue your contributions to MT.

I am happy too I found some time to improve the nutrition feature.

> I've not yet looked into the code but can the new fields in the tooltip be optional like other fields?

I will look into it.

> I'm waiting with this merge until MT 25.4 is published and this will be after eastern.

Sure. Did you mean "Easter" ?

FJBDev:
> I've not yet looked into the code but can the new fields in the tooltip be optional like other fields?

I have added the ability to hide the fields like the other fields

image

bundles/net.tourbook/src/net/tourbook/nutrition/DialogCustomTourNutritionProduct.java
```diff
@@ -185,6 +230,10 @@ private void createUI(final Composite parent) {
                   .hint(_pc.convertWidthInCharsToPixels(5), SWT.DEFAULT)
                   .align(SWT.BEGINNING, SWT.CENTER)
                   .applyTo(_spinnerBeverageQuantity);
+            if (_isEditMode) {
+               _spinnerBeverageQuantity.setEnabled(_product.isBeverage());
```
wolfgang-ch, line 234: I'm putting all control enablements into one place, e.g. `enableControls()` which makes it also easier to find issues

wolfgang-ch (commented):
I've found 2 issues which can make the code better maintainable

bundles/net.tourbook/src/net/tourbook/nutrition/DialogCustomTourNutritionProduct.java
FJBDev: Fixed

bundles/net.tourbook/src/net/tourbook/nutrition/DialogCustomTourNutritionProduct.java
```diff
@@ -109,6 +122,9 @@ private void createUI(final Composite parent) {
                   .span(2, 1)
                   .align(SWT.BEGINNING, SWT.CENTER)
                   .applyTo(_txtName);
+            if (_isEditMode) {
+               _txtName.setText(_product.getName());
+            }
```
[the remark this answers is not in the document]
FJBDev: Fixed

bundles/net.tourbook/src/net/tourbook/nutrition/DialogCustomTourNutritionProduct.java
```diff
@@ -102,6 +102,8 @@ protected Control createDialogArea(final Composite parent) {
       }
       _isInUIInit = false;
 
+      updateUI();
+
```
wolfgang-ch: Should `updateUI()` not within `_isInUIInit = true/false`. I think `_isInUIInit` is only need when setting field contents to prevent that modify listener are doing their work

wolfgang-ch: The flag `_isIn...` is needed for some controls and events when the content of controls is modified programmatically and not by the user to prevent that modify events are not performed when they are fired. 

For some controls / events it is not needed but after 20 years of MT programming I cannot remember in my brain when it is needed and when not, so for me it is always a try and error.

bundles/net.tourbook.web/WebContent-rel/tourbook/package.json
```diff
@@ -5,5 +5,5 @@
 	"description":	"MyTourbook web applications.",
 	"homepage":		"http://mytourbook.sourceforge.net",
 	
-	"dojoBuild":	"build-profile.js",
+	"dojoBuild":	"build-profile.js"
```
wolfgang-ch, line 8: I've no idea what this is fixing but the ant `net.tourbook.web CREATE-Dojo-Bundle.xml` is still running without problems

FJBDev: Theoretically, the extra `,` is not allowed and the build was complaining about. Now it should be good. I am also not familiar a lot with Dojo

PR #1520, opened 2025-07-19
Swimming FIT Unit test fix

FJBDev:
@wolfgang-ch A note should be added in the 25.next readme to inform the users that swimming tours previously imported with a FIT file should be reimported.


wolfgang-ch:
> @wolfgang-ch A note should be added in the 25.next readme to inform the users that swimming tours previously imported with a FIT file should be reimported.

Done https://github.com/mytourbook/mytourbook/commit/6621570b4ee1164631d0a58e7faca2046476671a

FJBDev:
And when I click on the commit hash, it shows this message

image

wolfgang-ch:
@FJBDev I replaced my commit with your pull request by using cherry picking, your commit is now also in the git history execpt that the sequence is not the same but this doesn't matter in this case

PR #1531, opened 2025-07-29
25.8 French translation, Java 21 for Github Build, UI test fixes and  Switching to the v3 Suunto Workout API

FJBDev:
- Replaced regular ASCII space (U+0020) between the time and "AM with narrow no-break space (U+202F) as it was making the UI tests fail recently. Maybe due to moving to the embedded Edge browser ?


wolfgang-ch (approved):
Nothing to complain 😄

PR #1558, opened 2025-08-25
Partial unit test fixes, MT system & build updates (Gradle 9, package updates)

FJBDev:


build/build.net.tourbook/build.gradle
```diff
@@ -159,33 +152,26 @@ task mt_40_ExtractWinFiles {
 /*
  Create win exe files
 */
-task mt_50_CreateExeFiles {
-        
-   doLast {
-       
-      println 'Creating Windows setup files'
+task mt_50_CreateExeFiles(type: Exec) {
```
wolfgang-ch, line 155: Only one windows installer is created instead of two

image


The build process runs without any errors

image

wolfgang-ch (changes requested):

bundles/net.tourbook/src/net/tourbook/importdata/RawDataManager.java
```diff
@@ -4186,7 +4186,7 @@ public void updateTourData_InImportView_FromDb(final IProgressMonitor monitor) {
 
                new ProgressMonitorDialog(Display.getDefault().getActiveShell()).run(
                      true, // fork
-                     false, // cancelable
+                     false, // cancellable
```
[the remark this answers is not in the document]
FJBDev: My bad, I learned something new today 😄. Reverted

build/build.net.tourbook/build.gradle
FJBDev: I will try locally, it works on the GitHub build

FJBDev: I had to switch machine recently for my MT development and need to install maven etc.... I will let you know once that is done

FJBDev: My build fails with the error below. Any idea why ?

image

wolfgang-ch: There is a local and reduced git repo to speedup the build process. The startingpoint is this readme file https://github.com/mytourbook/mytourbook/blob/main/build/build.eclipse.4.26.nl.repo/How%20to%20update%20this%20p2%20repo.txt

wolfgang-ch: The build do not run, it creates these logs

[mt-build.log](https://github.com/user-attachments/files/22229883/mt-build.log)
[log.txt](https://github.com/user-attachments/files/22229886/log.txt)

wolfgang-ch: In your screenshot, the `.exe` files are much larger than when I created it. Because the build do not run, so I can't can say how large my `.exe` files will be

wolfgang-ch: ... and the gradle version

image

wolfgang-ch: With tycho 4.0.13 the build runs without an error and the .exe file are not much larger than before

[mt-build.log](https://github.com/user-attachments/files/22237757/mt-build.log)

image

FJBDev: > ... and the gradle version
> 
Yes, but I had already upgraded it the last time you tried to run the build.

>With tycho 4.0.13 the build runs without an error and the .exe file are not much larger than before

I can downgrade the tycho build if you prefer but it runs fine on my Windows and on the GutHub build

FJBDev: >With tycho 4.0.13 the build runs without an error and the .exe file are not much larger than before

Can you try with Tycho 4 to see if Tycho 5 is actually the culprit ? I can't reproduce it locally, it works fine for me

wolfgang-ch: > > With tycho 4.0.13 the build runs without an error and the .exe file are not much larger than before
> 
> Can you try with Tycho 4 to see if Tycho 5 is actually the culprit ? I can't reproduce it locally, it works fine for me

This is what I already did.

I found the same error in https://github.com/eclipse-tycho/tycho/issues/4226 with the hint that a newer maven version can fix it, and it fixed this issue. Now the build with tycho 5 is working 😄

PR #1564, opened 2025-09-05
Add first draft of preference page for bikes

cyc1ingsir:
Related discussion can be found at https://github.com/mytourbook/mytourbook/discussions/1563

### DIFF
--- a/bundles/net.tourbook/src/net/tourbook/data/TourBike.java
+++ b/bundles/net.tourbook/src/net/tourbook/data/TourBike.java
@@ -15,7 +15,9 @@
  *******************************************************************************/
 package net.tourbook.data;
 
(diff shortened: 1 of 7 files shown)


bundles/net.tourbook/src/net/tourbook/data/TourBike.java
```diff
@@ -53,9 +77,6 @@ public class TourBike implements Serializable {
     */
    private int               typeId;
```
wolfgang-ch, line 78: A type should not be an integer value, this is difficult to read, it should be an enum (string) value

wolfgang-ch (changes requested):
this is not a complete review but I added a few comments which I found so far after a short review

PR #1685, opened 2026-05-03
Tour Nutrition fixes and new feature (tour nutrition filter)

FJBDev:


bundles/net.tourbook/src/net/tourbook/tour/filter/TourFilterManager.java
```diff
@@ -1565,6 +1588,28 @@ private static void getSQL_ManualTour(final StringBuilder sqlWhere,
...
+      if (fieldOperator == TourFilterFieldOperator.IS_AVAILABLE) {
+
+         operator = "IN";//$NON-NLS-1$
+
+      } else {
+
+         operator = "NOT IN";//$NON-NLS-1$
+      }
+
+      final String whereClause = String.format(
```
wolfgang-ch, line 1606: I would use this statement to format strings 

`"TOURID %s (SELECT TOURDATA_TOURID FROM %s)".formatted(...)`

-> less code and better readable

FJBDev: What is the benefit since I already use `String.format()` ?

wolfgang-ch: For me it is less writing

PR #1690, opened 2026-05-19
Tour Nutrition View: Update button is now in the contextual menu

FJBDev:


wolfgang-ch:
Currently there is no response in the UI when a product update was performed. I think it is more userfriendly when there is response and the simplest way is to show a message in the statusline with `net.tourbook.common.UI.showStatusLineMessage(String)` which I use e.g. when something is copied into the clipboard

FJBDev:
> Currently there is no response in the UI when a product update was performed. I think it is more userfriendly when there is response and the simplest way is to show a message in the statusline with `net.tourbook.common.UI.showStatusLineMessage(String)` which I use e.g. when something is copied into the clipboard

Yes, when a product is updated, the log view is opened and the differences are displayed

image

wolfgang-ch:
Because I've set the show log to `Never`, so I could not see it. I just created a feature request to make log changes better visible https://github.com/mytourbook/mytourbook/issues/1691

FJBDev:
> Because I've set the show log to `Never`, so I could not see it. I just created a feature request to make log changes better visible #1691

If you want to make log changes better visible, just check again the checkbox to see the log view. That's easier. I don't understand the need to create another place to see log changes, it seems overkill to me

wolfgang-ch:
> > Because I've set the show log to `Never`, so I could not see it. I just created a feature request to make log changes better visible #1691
> 
> If you want to make log changes better visible, just check again the checkbox to see the log view. That's easier. I don't understand the need to create another place to see log changes, it seems overkill to me

I would like to see log changes even when the log view is closed. 

I'll will implement it and I think that is not much work compared with what I'm currently fighting: the sql to filter out tours which do not have an equipment part

bundles/net.tourbook.cloud/src/net/tourbook/cloud/dropbox/DropboxClient.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (C) 2020, 2026 Frédéric Bard
+ * Copyright (C) 2020, 2024 Frédéric Bard
```
wolfgang-ch, line 2: Back to the future?

FJBDev: Yes, I had to revert this recent change

image

PR #1740, opened 2026-08-24
Tour Nutrition View: Fix for the recent products

FJBDev:
- Unit tests fixes
- Removed the Suunto 9 import (My very first MT feature back in 2018 !)
- Disabled the Strava upload and initial connection feature


wolfgang-ch:
>     * Removed the Suunto 9 import (My very first MT feature back in 2018 !)

What was the reason why you have removed the Suunto 9 import? It doesn't concern me but such devices can still be available.

FJBDev:
> > ```
> > * Removed the Suunto 9 import (My very first MT feature back in 2018 !)
> > ```
> 
> What was the reason why you have removed the Suunto 9 import? It doesn't concern me but such devices can still be available.

Because Suunto doesn't offer to access directly the files anymore when connecting the watch and since the Suunto API is available since several years now

PR #1748, opened 2026-09-04
Adding the average carbohydrates per hour in the tour book and calendar views

FJBDev:
- Removed the feature "When encountering invalid files, those files are marked to be ignored so
that they are not considered as files to be imported in the future" as the Suunto 9 import was recently removed [here](https://github.com/mytourbook/mytourbook/pull/1740)


bundles/net.tourbook/src/net/tourbook/data/TourData.java
```diff
@@ -2204,6 +2209,7 @@ public TourData() {}
    public void addNutritionProduct(final TourNutritionProduct nutritionProduct) {
 
       tourNutritionProducts.add(nutritionProduct);
+      this.computeTourNutritionData();
```
wolfgang-ch: I think that the code is better readable that `this.` is used only, when needed

`this.computeTourNutritionData();` --> `computeTourNutritionData();`

There are also other locations with `this.computeTourNutritionData();`

bundles/net.tourbook/src/net/tourbook/ui/views/tourBook/TVITourBookItem.java
wolfgang-ch, line 244: `SQL_ALL_OTHER_FIELDS__COLUMN_START_NUMBER ` is not adjusted which is **!!! EXTREMELY IMPORTANT !!!**

info/release-notes/26.6.0-readme.txt
```diff
@@ -18,7 +18,7 @@ New / Improvements
   - Select a product from a list of the most recent products used
 
 * Equipment
-  - Add a contextual menu to add several equiments at a time
```
wolfgang-ch, line 21: equipment has no plural, `equipments` -> `equipment`

wolfgang-ch (commented):
a few minor issues but the field `SQL_ALL_OTHER_FIELDS__COLUMN_START_NUMBER` is important, to be correct

info/release-notes/26.6.0-readme.txt
FJBDev: I found a lot more occurrences with equipment**s** and fixed them

bundles/net.tourbook/src/net/tourbook/data/TourData.java
FJBDev: Fixed

bundles/net.tourbook/src/net/tourbook/ui/views/tourBook/TVITourBookItem.java
FJBDev: Good catch. Fixed

bundles/net.tourbook/src/net/tourbook/database/TourDataUpdate_062_to_063.java
```diff
@@ -0,0 +1,61 @@
...
+   @Override
+   public boolean updateTourData(final TourData tourData) {
+
+      final IPath stateLocation =
+            Platform.getStateLocation(CommonActivator.getDefault().getBundle());
+      final File invalidFiles =
+            stateLocation.append("invalidfiles_to_ignore.txt").toFile(); //$NON-NLS-1$
+      if (invalidFiles.exists()) {
+         FileUtils.deleteIfExists(invalidFiles.toPath());
+      }
```
wolfgang-ch, line 51: Deleting the old file is a good idea but this code is called for each tour, in my case more than 5000 times, a better location would be in `getTourIDs()` which is called only once.

The constructor is also not a good idea because this would skip the version check.

I'm also not sure if the concurrent execution of this code could cause problems when run in `updateTourData()`

wolfgang-ch (commented):
Now the wording `equipment` is 100% perfect 👍 

The db update could be optimized

bundles/net.tourbook/src/net/tourbook/database/TourDataUpdate_062_to_063.java
FJBDev: I thought about that too as when I was debugging I saw the code being called for every single tour.
I've update the code to perform the file deletion inside the `getTourIDs()`

bundles/net.tourbook/src/net/tourbook/database/TourDataUpdate_062_to_063.java
```diff
@@ -0,0 +1,59 @@
...
+   @Override
+   public List<Long> getTourIDs() {
+
+      final IPath stateLocation =
+            Platform.getStateLocation(CommonActivator.getDefault().getBundle());
+      final File invalidFiles =
+            stateLocation.append("invalidfiles_to_ignore.txt").toFile(); //$NON-NLS-1$
+      FileUtils.deleteIfExists(invalidFiles.toPath());
+
+      return List.of();
```
wolfgang-ch, line 45: When an empty tour id list is returned, then the method `updateTourData()` is run for 0 tours.

I'll fix this by returning `null` that I can merge this PR

image

bundles/net.tourbook/src/net/tourbook/database/TourDataUpdate_062_to_063.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.Platform;
+
+public class TourDataUpdate_062_to_063 implements ITourDataUpdate {
+
+   @Override
+   public int getDatabaseVersion() {
+
+      return 62;
```
wolfgang-ch, line 33: I just detected, that the version number 62 is wrong, it should be 63.

I'll fix it.

bundles/net.tourbook/src/net/tourbook/database/TourDatabase.java
```diff
@@ -11966,6 +11978,30 @@ private int updateDb_061_To_062(final Connection conn, final SplashManager splas
       return newDbVersion;
    }
 
+   private int updateDb_062_To_063(final Connection conn, final SplashManager splashManager) throws SQLException {
+
+      final int newDbVersion = 62;
```
wolfgang-ch, line 11983: I just discovered and changed this wrong db version: 62 ->63

FJBDev: Good catch and I saw you fixed it in `main`

bundles/net.tourbook/src/net/tourbook/database/TourDataUpdate_062_to_063.java
FJBDev: Good catch and I saw you fixed it in `main`

bundles/net.tourbook/src/net/tourbook/database/TourDataUpdate_062_to_063.java
FJBDev: Oh I missed that. It worked that i tested it locally

